Issue #6, opened 2017-03-15
Communist South Africa does not end apartheid


ARR8:
On 1.05, AI South Africa consistently takes a decision to make Black Anglo the dominant culture (majority rule, cpc_SAF) upon regime change yet keeps the apartheid modifier and the Bantustans.
Will submit possible fix.

Additionally, black anglo does not seem to be very common in South Africa, maybe "Coloured" should be added as accepted?

settintotrieste:
Thanks for the help! I will add it to the mod.

ARR8:
No problem!
On second thought, looking at the names and category for the Coloured culture, it seems like a stand-in for a generic non-tribal South African, so I would personally leave them to be assimilated. Your call, of course.

settintotrieste:
Thanks, for now I will leave coloured to be assimilated.

PR #7, opened 2017-03-15
Allow ending bantustans if black anglo dominant #6

ARR8:
Observed behavior: AI South Africa ends Bantustans if communist, does nothing if not.
Ending apartheid this way does not relinquish the core on Walvis Bay if Namibia exists, as the original method does. I'm not sure if it should, since it is uncertain whether Namibia will exist, and there is a decision just to return it. In the end, the only difference should be the amount of infamy lost.

### DIFF
--- a/CWE/decisions/south_africa.txt
+++ b/CWE/decisions/south_africa.txt
@@ -132,16 +132,20 @@ political_decisions = {
 				exists = VED
 				exists = ZUL
 			}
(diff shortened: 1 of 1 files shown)


ARR8:
No problem! I've got a bunch more bugs to go lol. Would you rather I have a shot at fixing them first, or just submit all the issues?

Issue #12, opened 2017-03-15
Arabia takes Salafism even if atheist


ARR8:
v1.05, title. May want to set AI will do to 0 if party policy is atheist or secular, and/or if secularism reform is enacted.

It seems that most of the Arabian parties are secular, and I'm not sure about the logic of the reform or the decision which changes it.

settintotrieste:
Thanks, that was a good catch! I will add your fix.

PR #13, opened 2017-03-15
Arabia will not enact wahhabism if atheist #12

ARR8:
This ignores the status of the "religious rights" reform. It may be better to use its status in conjunction with party policy, especially as it seems to be more or less fluid.


ARR8:
Whoops, put the issue number in the wrong title. Don't think it matters that much.

PR #23, opened 2017-03-17
Leader portrait enhancement

echasovikov:
Well, the current state of the general/admiral portraits is a mess, with only a limited number of modern age portraits available, and most of the portraits used being vanilla ones. So, I replaced all of them with photos of modern/late cold war generals.


settintotrieste:
This is truly excellent! Thank you very much!

PR #26, opened 2017-03-17
Yugoslav ethnic cleansing fix

ARR8:
Prevent Croatia, Bosnia, etc. from ethnic cleansing when they are puppets of Yugoslavia as part of the 1974 constitution event.
Maybe this is better as part of potential rather than allow? Seems like a player interface question.


settintotrieste:
I think having it under allow is better because communist dictatorships can be allowed to conduct genocide but only independent dictatorships can actually do it.

Issue #21, opened 2017-03-17
Province mistakes


settintotrieste:
Someone said:

Some of the regions seem to be a bit off. Tobolsk (in eastern Siberia), for example, is split entirely in two by Upper Yeniseysk. Furthermore, Lower Yeniseysk and North and Middle Siberia have made their way to eastern India.

ARR8:
Furthermore, in my test run, it seems that all of the Indian states are colonial except for the capital state and Hyderabad. More testing needs to be done to see if this happens every time, but the AI does not promote bureaucrats enough to be able to make them states on its own if this happens.

settintotrieste:
Thanks for the feedback however for now it is necessary for most of the Indian states to be colonial because India's economy will grow extremely unrealistically quickly, usually no. 2 by 1992 when they have all provinces as non colonial. A possible fix for this is to specially dramatically reduce promotion to workers for India so that the speed of industrialisation can be slowed down to more realistic levels.

ARR8:
Ah, that makes sense. In that case, maybe Hyderabad should be addressed so that it is set to colonial after it is conquered, so it doesn't get massively industrialized and become a metropolis. I'm not sure how you would do that, maybe setting the Hyderabad tag to uncivilized? But then the war would be too unbalanced between them.

EmperorRAG:
To make it colonial. You release it as a vassal and set it as uncivilized, then annex it.

ARR8:
Thanks for the suggestion, @EmperorRAG

settintotrieste:
I think there may be a problem with making Hyderabad a non colonial state. Currently as the Indian AI is incapable of creating enough bueacracy to promote other colonial states to states, only Delhi and the Kashmir and Hyderabad are states in the whole of India. Making Hyderabad a state may mess with their current economic growth pattern which allows their industrial score to reach at least 12th by 1992. So I will probably not want to put this into the mod as of now..

ARR8:
@settintotrieste, if I understand what you're saying, then you want to keep three Indian states as non-colonial so it gets a reasonable industrial score about when it should. If that is what you mean, that makes sense, and I agree. However, I can see a couple of problems with this method:
1. It keeps India too dependent on keeping Kashmir, as one third of its potential industry. In my games, China goes to war for the whole state relatively often.
[...]

PR #34, opened 2017-03-19
Make Hyderabad colonial upon annex #21 + cores

ARR8:
This method changes the CB to a cut down to size, which, I hope, shouldn't mess with the AI's willingness to surrender. Nevertheless, there is such a power difference between Hyderabad and India that it shouldn't matter.


settintotrieste:
I think there may be a problem with making Hyderabad a non colonial state. Currently as the Indian AI is incapable of creating enough bueacracy to promote other colonial states to states, only Delhi and the Kashmir and Hyderabad are states in the whole of India. Making Hyderabad a state may mess with their current economic growth pattern which allows their industrial score to reach at least 12th by 1992. So I will probably not want to put this into the mod as of now..

PR #40, opened 2017-03-20
Allow India to begin industrializating Bombay and Madras #21

ARR8:
Someone other than me may want to double-check the trigger to make sure this doesn't happen when it's not supposed to.
  Kashmir still needs to be made colonial. We have two methods of making Hyderabad colonial, one by release and integration as in ecc883c and one by editing the war as in #34.
  For industrializing Bombay and Madras, we can use this method, or set them as states from the start of the game under the British Raj, just like the capital state (United Provinces).

There are pros and cons to each of the options. Personally, I would go for the ones with fewer events, as you almost get spammed with them starting as India. However, any choice will work equally fine for setting up a more reasonable industrialization of India, which was the goal.


settintotrieste:
Thanks however I think the setting Bombay and madras as states from the beginning is a better option. It would involve less events that may cause problems during ahistorical situations plus having less spam. So I will set madras and Bombay as states in the 1946 start.

PR #45, opened 2017-03-20
Special Economic Zones - fix and enhancement

echasovikov:
Both the decisions to establish the zones and the event to end the special economic zone participation didn't work previously. They do now.

Moreover, participation in foreign Special Economic Zones requires not government form and political ideology, but free trade and either laissez-faire or interventionism policies.


settintotrieste:
Thank you for the fix however I find that ability to engage/start in the SPEZ program should be tied to government type to prevent swamping of the world of SPEZs or too little due to too many program recipients.

Issue #18, opened 2017-03-17
Non state actors: Terrorists


settintotrieste:
Main feature for 1.06

settintotrieste:
Feature added however tuning of frequency of attacks is still required.

PR #46, opened 2017-03-20
Terrorism rewrite #18

ARR8:
All terrorism events are selected from at random by one MTTH event, to simplify tweaking of terrorism frequency. This comes with a loss, namely that different terrorism events cannot have different requirements. However, the only requirement used, that of a minority discrimination policy, I felt, is out of place. The attacks allowed only to happen in countries with that policy have happened in real life in countries one would not think of as having institutionalized discrimination as a law e.g. the recent mosque attack in Canada and the assassination of Bobby Kennedy in America.
  The exception to this is the ocean liner hijacking, as I could not select coastal provinces. I have left this as a MTTH event for now.

I would like to add more unique consequences to each event, such as the one I added for the hostage crisis. I guess that would be a TODO. 

For now, I have not changed the MTTH modifiers, except for adapting them to the province scope so that the event will fire.

#### Playtesting is needed! I only tested to see if the events will fire!

I would also like to make the terrorist attack event a major one, but my description is too long. I will probably shorten it soon and test.

I would welcome any feedback!


settintotrieste:
Thank you for the contribution however I prefer not to change the terrorism events such that there would be a generic event to select the actual event that would occur as it would be rather clunky and strange. I prefer it to be like the base Victoria province events where each event happens on the basis of the province conditions. But what I will do is to remove the minorities oppression criterion and implement the code fixes that you have done.

ARR8:
I see what you mean, but I would strongly urge you to reconsider. Currently, if anyone wants to tweak the MTTH of the event, they have to make the same change nine times. Adding in each of the unique forms of terrorism seen in the mod's timeframe, it could easily double that number. Even the current amount of code reuse can get frustrating very easily, not to mention the file size will multiply each time.
  Having it go through one event would allow addition of a new type of terrorism with five lines, the probability of the event changed with one number, and then custom options for the events added in a clear way.

Plus, the description on the general event outlines the consequences of terrorism and the ways to prevent it. If this is could be a major feature, which I think would be good as currently Vic2 lacks challenging mechanics for experienced players, this needs to be explained somewhere, and currently reading the tooltips on a modifier is a bit obtuse.

**However**, there is another way that could be considered that sort of combines the benefits of both. There could be a general event, like in this PR, that fires on complex conditions and does nothing but add a province flag/modifier. Then, each of the terrorism events is a very low MTTH event that triggers on provinces with the flag/modifier. This would allow unique events based on province conditions, and would still make it much easier to develop: only one copy of the massive MTTH block, and it would still be possible to edit the probability of each terrorist event.

Let me know if that would be better or you want to see an implementation and I'll submit a new PR. If you want to keep the current system, I understand, but I probably will not try to add terror events until the frequency is fully balanced, whenever that happens.

Thanks for considering!

settintotrieste:
Thanks, I see the problem now and your solution is a good way around it so going along the lines of your proposal, there can be a generic event with the complex conditions called "fears of a terrorist strike" to set the province flag terror_strike_impending and then the actual terrorist events with a very low MTTH but with simple conditions like is port or is state capital etc to allow differentiation of events. The actual terrorism events will then add the province modifier of terrorist_chaos and the country modifier of recent_terrorist_attack and clear the province modifier added in the generic event. If you can implement this in the new PR, it will be excellent!

ARR8:
@settintotrieste - I have implemented and tested the system described. Please see f1f9f08. I have also began to work towards issue #50 with commit da7c6b6. Please let me know what you think about either.

settintotrieste:
Thank you, this is excellent!

Issue #50, opened 2017-03-21
Socialism too popular in Europe and late-game


ARR8:
[/u/EmperorPeriwinkle said:](https://www.reddit.com/r/paradoxplaza/comments/60hsio/a_few_things_about_the_cold_war_enhancement/)

Speaking of livening up elections, I'm at 2020, UK has been lead by the Labour party for all but a couple years in the 50s. France has being ruled non stop by the socialist party. The CDU has only beaten the SPD once and that's because of the Spiegel Affair event tanking them with a ruling party scandal modifier. Ironically the modifier stayed after the election and the event chain got to the point an election was triggered 2 years into CDU rule and the SPD won with a straight up PR majority. From that point the SPD may have lost 2 more elections, max, were it for my use of espionage. Once Die Linke became a party SPD has started alternating leadership with them, but non socialists aint ever winning in Germany. Socialists are straight OP the western world. You know how the centrist moderate monopoly in USA died? The Socialists winning every time between 2000 and 2020. I'm sure this is happening because blue collar worker Pops are massively socialist and the way industrialization works, it means socialists only get stronger as the game goes on. Also meeting needs is super easy for western countries, which increases the ideology support of the ruling party.

-----------

I will be submitting a few pull requests to do with this and #18 to make pops more right-wing in the face of terror attacks on the country as a partial solution.

settintotrieste:
That's a good idea to have the terrorism events shift the pops to the right.

I have done some tweaks to reduce socialist support among craftsmen so as to allow other parties to have a greater chance of winning the elections. More testing still needs to be done though.

settintotrieste:
More dynamic ideological alignment attempts to fix this.  aef0e33

PR #54, opened 2017-03-21
Add cooldown to national value change

ARR8:
I have noticed that the AI changes the country's national value whenever the pop majority ideology changes. However, this can change very quickly, especially for small nations where one event can add support for an issue for the whole country.

It seems to me that a fundamental shift in the important value of a country should have a cooldown. Ideally this would be generational, but given the abstract nature of the game, I set a preliminary value for four years, long enough for the political discourse of a country to change. I have also changed the text to give an explanation of what changing the national value would entail.
  The cooldown uses the modifier used when switching tech schools, to block a nation from changing both, and the effects make sense, in my opinion.

To help issue #50, I have allowed countries to switch to order as a response to terrorism, to simulate the rise of the right in our day.

If you like these changes, do not merge the PR yet, but let me know and I will also add conditions for AI_will_do to all of these, to prevent, for example, dictatorships changing their national value to liberty.


settintotrieste:
Thanks but I might not want to add this in because the national value should represent what the majority of people were subscribing to at any one point. If the majority of people suddenly start becoming liberal then the national value should also change to liberty by the AI. I think adding the academic_confusion modifier will unnecessarily make the changing of values a punishment for 4 years. 

A compromise solution would be to allow the change into the respective national values only if the government were subscribed to the majority ideology as well. So in a democracy the national value would still more or less follow the majority ideology but not as fast as the necessary party still needs to be elected. While in dictatorships the default will be the order value and in Islamic states, Islamic NV.

ARR8:
Okay. I've always interpreted the national value as being the type of discourse and solutions the culture and history of a country tends to promote, so a pop, given all other factors equal, would be more likely to support a certain ideology in one country than another because of how he was raised etc.

So I'm not sure it should change at all, sort of like how the Islamic NV is implemented now, except for a dramatic shift in a country's direction, like the War on Terror. I actually wasn't a fan of how NWO added industrialization as an NV rather than as a modifier because it says nothing about political culture.

In terms of gameplay, I'm not sure it makes sense for the NV to represent what the majority of people think, since it would just encourage more of them to think that way, and then everyone would end up with an equality or order NV and just having the NV would make it stay that way. Coming from the perspective above, I think it should shape pop opinions, not be shaped by them, and make it easier to return to the baseline type of politics in that country after whatever happened.

In either case, whether the NV represents the "national ideas" of a country or the current political climate of its pops, I think we should add least change the localization for the decisions to explain what is actually being changed in equivalent real-world terms so the player can decide if it makes sense to do, instead of just thinking about gameplay implications.

And I agree completely on the government affecting AI_will_do. That was what I was going to do next.

settintotrieste:
I understand what you're saying however there is a problem with making the national value stay the same most of the time as many of the national values set for many of the countries in the base NWO that I used are just equality. So this was kind of a compromise solution to allow the countries to dynamically switch their national values along the way to simulate the changing values of nations throughout the 100 years of the game. An example would be the UK in real life where the national focus went from valuing order (empire days) to equality (1960-70s) to meritocracy (post 80s) then perhaps equality again today. 

But in the meantime you can tie the change in government to change in NV and do the changes in the localisation.

ARR8:
Hm, that makes sense. I feel that we're missing whatever it was that really caused those changes so those examples would never happen in a normal game, One could argue that there is always a streak of order in the UK, such as was revealed in the Brexit vote, despite UKIP remaining a minority in elections, and that equality/meritocracy weren't changes in value but just the political climate. For example, the typical aristocratic Englishman who values conservatism and the old guard is still more alive in the UK than many other European countries. In this case, order is still the NV, just pops prioritize different things.

I guess this would be an instance of leaving it how it is or removing it and coming back to it once shifts in politics are implemented in more detail, and tying it in somehow.

I will edit the PR to just include government and changes in localization, and make a new issue for political shifts.

Thanks for the discussion!

settintotrieste:
Thanks! However this argument applies primarily to the rich upper class which almost always support order anyway as a means to protect their status and not really representative of the innate "national alignment". There will be also be a problem in interpretation when it comes to deciding what is the innate "national alignment" of a nation and doing that for all the nation's in the game will be painstaking so it is better to shift the interpretation of "national value" to what is the "national focus" of the time for pragmatism. Furthermore there is a difficulty in gauging what should constitute a "seismic event" beyond the scripted ones to shift the "national alignment" and generic election events definitely won't cut it.

PR #59, opened 2017-03-23
International media now reports on terror + fixes

ARR8:
Whenever a terrorist attack happens and is not prevented (possibly TODO mechanic, for now only if hostages are rescued), the far-right will gain ground and demand changes. Whether the country accedes or not, the far right movement nationwide and in a couple of other countries with a free press will pick up, because it is reasonable to observe that terror attacks in other countries still scare people, especially in the West.

The changes are slight but frequent terror attacks will make them add up. Either way, for a brief period of time, the far right movement is brought into the national consciousness...

Fixes include scopes and insulating rich people from some of the chaos.

### DIFF
--- a/CWE/events/Terrorism.txt
+++ b/CWE/events/Terrorism.txt
@@ -452,8 +452,8 @@ province_event = {
 	
 	option = {
 		name = EVT_1511119_A
-		prestige = -5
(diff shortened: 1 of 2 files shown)


settintotrieste:
Thanks, this is great!

PR #63, opened 2017-03-24
Terrorism organization, conditions, effects, event

ARR8:
By the way, the reason I made the terror attacks major events is that countries would have some context if they got event 1511121 (National security called into question). I can see the point in removing it if there are going to be a lot of attacks, but then maybe that event might confuse a new player - something to consider.

### DIFF
--- a/CWE/events/Terrorism.txt
+++ b/CWE/events/Terrorism.txt
@@ -259,32 +259,26 @@ province_event = {
 		months = 60
 
 		#Increase - Unemployment
-
 		modifier = { 
 			factor = 0.9
 			unemployment = 0.1
 		}
-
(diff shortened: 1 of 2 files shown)


settintotrieste:
I think it would be better not to have it as major events to prevent possible spam of events for other countries because there could be many attacks happening simultaneously especially towards the endgame.

ARR8:
Okay. Do you plan to ever put an end to the "war on terror" age, or leave it until the end date?

settintotrieste:
Yes, I intend to remove it by the end game but I'm not sure what is the best way to do it yet. There is no real world precedent on how to diminish the global terrorist threat.

PR #69, opened 2017-03-25
Prevent terrorism during decolonization

ARR8:
Starting a new game, terrorism events were more frequent than all other events, especially as a colonial power who already gets colonial unrest events. Although it is good to make them less frequent after a recent attack as in 040c049, it seems that there shouldn't be an attack during the 40s at all.

This also makes them half as frequent overall, in addition to year-based tweaks. Also prevents them for colonies.


settintotrieste:
Thanks

PR #71, opened 2017-03-25
Fix Manchuria infamy cheat for USSR

ARR8:
Currently, the USSR can release Manchuria to prevent the event asking for return to China, or wait for the event and release it before picking an option for a free infamy reduction. This prevents that, and adds an option for a Mongolia-style puppet state for slightly less infamy than to keep it if the player should so choose; the AI never will.


ARR8:
By the way, there is a similar problem for Austria that I would like to fix. It is a little more complicated since it involves other occupiers, however.

settintotrieste:
Thanks

ARR8:
No problem - one more thing, in this change, Port Arthur gets returned to China even if Manchuria gets released. Let me know if you think it makes more sense to change that. Personally, I imagine that the USSR forces Manchuria to remove their claim on Port Arthur, but the Chinese aren't held to that.

settintotrieste:
Sure this is a plausible scenario as the USSR did intend to return port Arthur to gain good will with a potentially valuable communist ally in the east and in real life port Arthur was fully returned to Chinese control in 1955 as a gift by Khrushchev to Mao.

ARR8:
Great - I'll leave it as is, then

PR #73, opened 2017-03-26
Fix #21

### LINKED ISSUE #21: Province mistakes

ARR8:
States were being overridden by vanilla darkness.csv. Moved into file higher in the alphabet.

I would recommend putting all the localization files into a CWE_ namespace to prevent this from happening again. I'll also be splitting the localization files over time up to organize them.


settintotrieste:
Thanks, I will take note of that.

PR #74, opened 2017-03-27
Change wartime rationing to set level with modifier triggered by war

ARR8:
This allows wartime rationing to be set once and then only activate the effects when at war.

This fixes a couple of problems for players with the current implementation:
- You have to take the decision to enact and end every war.
- There is no reason to take it off the lowest level after the war ends, since it only gives advantages.

It should not change AI behavior for rationing. Also tweaked localization.

If this is approved, I will do the same thing for martial law, which has the same problems.


settintotrieste:
Thank you however I prefer if it was not a triggered modifier as it would be difficult for the player to view the effects of the wartime rationing which they may need to guage what level of wartime law to use etc. What can be done instead is to add a negative modifier via an event to increase war exhaustion if rationing is not removed after a war.

ARR8:
I see what you're saying, but check again: I added all the effects of the modifiers to the decision descriptions.
  Even using that solution, it would still leave problem 1, and war exhaustion is already lowered by martial laws which have the same problem.

settintotrieste:
Sorry about that, I missed that out. Then in this case it should be fine. I will add it in and you can apply the same fix for the wartime law as well. Thanks

PR #75, opened 2017-03-27
#74 for wartime law

ARR8:


EmperorRAG:
With the Wartime Law decisions in place, could there possibly be a way in which it could be used to establish a dictatorship; such as through the use of the martial law policy and having a populist/communist/traditionalist/conservative party in power?

ARR8:
I think that's a good idea, but I'm worried it might be abused to get more player governmental control. It would have to carry some serious negative penalties and/or difficult requirements. If the former, the player could just never switch to the last level of wartime law, and if the latter, it would be so rarely seen by the AI that it would pretty much only happen intentionally.

I really like the idea, but some thought needs to be put into it.

PR #76, opened 2017-03-27
Superpower and Operation Mongoose rewrite

ARR8:
And strings. Just general cleaning, but should make things easier to change in the future.


settintotrieste:
Unfortunately I can't merge this in as the superpower modifier is used in multiple areas in the mod beyond the growing the superpower sphere files so removal of it will cause bugs.

ARR8:
Okay. I though I'd found and changed them all, though, and it seems they could all be replaced with the un_king modifiers.

In general I'd like to get around to reducing the total amount of modifiers because, a couple of decades in, there are way too many of them, and you can't even find and check the effects of any individual one. Some of them, even if maybe not this one, can definitely be safely merged.

settintotrieste:
I'm hesitant to remove modifiers though because they are quite useful for me to script new features on top of existing ones without needing to more code. I think it is better if we keep the superpower modifier as I might have some use for it in the Supepower join projects feature.

settintotrieste:
I think having too many modifiers is inevitable considering the complexity of the features in the game and the programming limitations of the Victoria code. I think what is the most important is to ensure that the decisions/events work well to minimise instability that can arise from such complexity. So I believe that it is not necessary to merge modifiers at this point.

ARR8:
Understood - I will keep it a low priority

settintotrieste:
Thanks

PR #80, opened 2017-03-31
Fixes and organization for Yemen CW and PER/SAA cold war

ARR8:
Still not tested, I only fixed the obvious problems I saw and made some enhancements.

Changes I'd still like like to see:
- Houthi rebellion before the civil war proper
- War outcome events only fire if intervention actually happened
- More dynamic relation changes, like I did a little for Persian intervention (relations down with all the African Arab states instead of individually Egypt, Morocco, etc.)
- High immigrant push for civil wars such as this and the Syrian one, simulate the refugee crisis
- Similar changes and fixes for the Syrian Civil War. There are still a lot of hardcoded elements that don't account for variability that I'd like to see become dynamic. That one is a little more complicated, of course.

And of course, testing needs to be done. If I had to guess, I would say Saudi involvement will immediately win the war. I tried to slow them down slightly with my second commit by lowering infrastructure, but I'm not sure how much it will help. It may be necessary to put a barrier between Arabia and Yemen such as the one that exists at the Himalayas.

### DIFF
--- a/CWE/events/iranian_saudi_arabian_cold_war.txt
+++ b/CWE/events/iranian_saudi_arabian_cold_war.txt
@@ -8,7 +8,9 @@ country_event = {
 
 	trigger = {
 		tag = SAA
-	        year = 2011
+		year = 2011 #consider replacing with appropriate cultural technology or flag, e.g. arab spring
+		NOT = { in_sphere = PER }
+		NOT = { is_sphere_leader_of = PER }
 		IRQ = { primary_culture = shia_arab }
 	}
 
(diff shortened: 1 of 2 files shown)


ARR8:
It turns out North Africa is part of the Middle East, not Africa, so that effect will need to be redone. Sorry!

settintotrieste:
Thank you very much for your help on this. May I know what is the progress so far? You mentioned something about effect needing to be redone.

ARR8:
I haven't really worked on it; the thing to redo is a minor effect changing relations with distant countries, but the whole method of changing relations should be rewritten to be more dynamic. I'll copy this list into the outstanding issue and do some elaboration.

Issue #77, opened 2017-03-28
Yom Kippur & Six day wars do not function correctly - Assigned to ARR8


settintotrieste:
From feedback:

The wars almost never breakout and Israel ends up stuck with 1948 territories.

ARR8:
I'll leave my other bugs with Israel here, too:

By the definition of the territorial dispute CB, Arab states can declare war on Israel. When Egypt declares a war, though, it messes with a lot of events that search for a war between EGY and ISR to do with the Suez crisis, triggering most of them early, and preventing the European powers from being involved.
  For example, the UN votes whether to intervene against Israel, even though they are the defender. Several European events fire, such as the FBU, referencing the Suez crisis which never happened. The result is that America intervenes to save a sphereling, but not an ally yet, and does nothing, leaving Israel fully occupied by Egypt for years.

Minor bugs:
Releasing Gaza before the partition can break future events such as the 1948 war and possibly others. Releasing Jerusalem, also before the partition, makes it quickly turn into an independent Palestine, which is realistic considering its demography, but it never gets the West Bank and then gets annexed to Jordan.

settintotrieste:
Thanks for the further feedback. This is more complicated a problem than I thought, I don't know how to approach this.

ARR8:
I should say; I'm working on rewriting all of the Israel events and adding more, but it will take a couple of days. If you want, you can assign the issue to me so it looks official :). But I won't be submitting any PRs until it's done. I'll be putting the changes as they come into a branch in my fork.

settintotrieste:
Sure and thank you very much for your help!

ARR8:
Ha, that's great but FYI there is a Github system for assignees, tags, and version milestones in the sidebar. That last one would probably be good to use to plan the main features of the next release and future ones.

settintotrieste:
Thanks for the heads up! Those will be very useful!

PR #84, opened 2017-04-01
Complete Israel rework

ARR8:
All of the existing events should be fixed to progress logically and actually happen, fixing #77. A couple of other ones were added to represent major territorial changes, and Israel can now fall. Additionally, everything has been relocalized.

Feel free to take your time approving this. There are a lot of changes, and, though I tested them all, it's possible I missed something.

There is still a lot to be done - I'm not sure whether to make a separate issue for all the things I want to add. Either way, they'll be on hold for the next PR because most of them will not create big changes on the map, and are mostly flavor for players.
  Some of the wars still don't work perfectly. For example, the AI is reluctant to join in the Arab-Israeli war. I'm not sure what can be done about that, though. The occupation of South Lebanon is greatly simplified, but, given that so is the Lebanese Civil War, I think it works well enough to get the right result.


settintotrieste:
Thank you very much! I have looked through the events and they are excellent and a huge step up from the NWO ones. This type of events are definitely the right step forward even if some of them don't work as well as it could be. I will add it to the master right away and look forward to your next PR.

PR #88, opened 2017-04-04
Fix land joining CSA in new civil war + optimization, strings

ARR8:


settintotrieste:
Please do not split the Manchurian candidate events from the essential decisions file. I need them in one file if the numbers are sequential from the last event in the essential decision 4 file to ensure I do not accidentally clash event ids.

ARR8:
Maybe if we just put a comment at the end of the ee4 file showing the ID range for the events? That would solve the problem and still keep things organized. You could also potentially reserve some event IDs for future Manchurian Candidate events, if you plan to add any.

Generally, from what I've seen, mods would create a new range for every new addition and differ them by 10, 50, or 100. For example, IDs 141300-141400 could be reserved for MC events, and the next hundred for the next thing. Most things probably won't even need 10 events, but it prevents having to shoehorn an event in a later addition using an unrelated ID.
  Vanilla does this, if you look at EventIDs.txt

Regardless, let me know if you want me to put e.g. the last one back in or leave a comment at the bottom.

settintotrieste:
I think it is better if you put the MC events back into ee4 as it was originally. I am more comfortable with such a layout.

ARR8:
I see. I'd rather not have commits just for moving things around, so I guess just merge this after you've put more events in the file and you don't need these in there anymore, and I'll keep it in mind for next time.

settintotrieste:
Thanks, I will merge it now then reallocate a new range of numbers to the MC events. So as to preserve the sequentiality of ee4.

ARR8:
Great solution - from now on I won't PR any of the essential events until they've been supplanted by another at the end of the file.

PR #93, opened 2017-04-06
Different newspaper titles for different countries

ARR8:
There are a lot of newspaper titles in the pictures, but none of them are being used. I took the news files for vanilla and adapted them to work with the mod. Some work still needs to be done. For example, not all of the images present in the mod are from vanilla countries; there are extra ones, listed at the bottom of the file, that need to be added.
  Additionally, vanilla counts on uncivs using the newspapers of their sphere leaders. This will have to be replaced; I suggest simply the sphere leader, as it would provide a cool, more visible effect. It can be arranged so that a country's own newspapers take precedence; this is only for those without it.
  Finally, there are images for countries with an anarcho_liberal ruling government. I'm not sure what the equivalent would be in the mod; I don't think there is one. Those can be kept for a future ideology, such as libertarianism, or removed.

### DIFF
--- a/CWE/news/news_layout.txt
+++ b/CWE/news/news_layout.txt
@@ -1,0 +1,1769 @@
+#How much printing tension gives every type of news.
+#(if not specified on the list, default value is 1)
+article_tensions =
+{
+	PEACE_OFFER_ACCEPT = 100
+	GAME_EVENT = 30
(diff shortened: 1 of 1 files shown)


settintotrieste:
This is excellent. Please go ahead. I think place anarcho-liberal under liberal ideology for now.

PR #98, opened 2017-04-09
UNSG tweaks

ARR8:
Optimization, strings, fixes.
  There are a couple small gameplay differences: there is much lower consciousness as a result of the UNSG selection, since a big jump in consciousness approx every 10 years seems like it could have consequences, plus I don't think that many people care about the UN so strongly;
  Any country can get the selection event, it is localized as though their candidate was chosen, and they get a small-medium amount of prestige. All nations are notified of the selection and country of origin.


settintotrieste:
There is a problem with the changes to the selection. As all nations can now have receive the UNSG selection event, there can be multiple clearing of global_flags simultaneously.

ARR8:
That's true, but it's very unlikely as the only event with a MTTH has a 0.003% chance of happening on a specific day the first time and a 0.00027% chance subsequently.
  Of course, that's in theory, and it's possible the MTTHs don't work like they're theoretically supposed to. Either way, that problem exists in the current implementation in the case of multiplayer games, so either one would have to be changed if turns out to be a real problem

settintotrieste:
I understand what you're saying but multiplayer support is of a very low priority for CWE as there are exceedingly few who do so. 

Yes however the mtth time will be applied separately for every nation that it affects therefore at around the end of each 120 month cycle, there can be a series of events firing by all the different nations that would rapidly change the Secretary General and spam the player with many notifications. So this is the best compromise solution. If it's possible you can modify the PR to include only the localisation changes.

ARR8:
Done - I'd like to find a more permanent fix, though, since I like the idea of a nation getting prestige from getting the UNSG seat. The prestige could even vary based on which of the three it is.

settintotrieste:
Thanks, though I need you to also remove the UN_ga member requirement for the selection event just in case the player is a vassal and is not in the GA. I like the idea of having your own nation's UNSG selected. I will try to implement it if it's possible.

ARR8:
Right, forgot about that one. I would recommend implementing it the way it was initially in the PR once a fix is found rather than integrating two sets of events, especially since the major event notification from the selection will notify the player which country the event is from.

settintotrieste:
Thanks! Yes it will be integrated into the current set of events. Probably just an extra event that is triggered by the UNSG selection event with a flat percentage of say 5% THIS provides UNSG and 95% a random country gets the event with the event a major to inform all nations.

ARR8:
Great - possibly the chance could be weighted by participation in UN events and international orgs, so you don't get populists taking pride in their UN leadership; but that's an issue for a different commit.

Issue #101, opened 2017-04-09
UN Secretary General Enhancement


settintotrieste:
Settintotrieste:

Probably just an extra event that is triggered by the UNSG selection event with a flat percentage of say 5% THIS provides UNSG and 95% a random country gets the event with the event a major to inform all nations.

ARR8:

Great - possibly the chance could be weighted by participation in UN events and international orgs, so you don't get populists taking pride in their UN leadership; but that's an issue for a different commit.

ARR8:
Using new mechanic from #102 TODO:
- Add a decision for populist or Manchurian Candidate nations to leave the UN, which will move the HQ if owned.
- Create an event that will move the HQ if the country is above the infamy limit - integrate into existing over-infamy event if there is one.

settintotrieste:
UN HQ can leave and resettle somewhere else if the country is nuked etc

PR #102, opened 2017-04-09
UN Headquarters mechanic + proposal for #101

ARR8:
This adds a small prestige ticking modifier to the nation with the UN Headquarters, and allows for its movement in some instances.
  Currently, this will happen in the case of a civil war from a Manchurian Candidate.
  **Oh, forgot to say: the HQ starts in NYC.

The UNSG events have been rewritten to trigger for only one country, the one which houses the UN headquarters. Then, the selection will move to a random UN member to simulate selection and all countries will be notified of the SG's character.

New UN localization file: Should eventually move all UN localization into this.


settintotrieste:
Great!

PR #105, opened 2017-04-14
Bloc logic redo and cleanup

ARR8:
Bloc modifiers unchanged, but moved into triggered_modifiers and activated by bloc flag.

The main gameplay change is the expansion of the NAM. Nations which are democratic/communist, but in the sphere of the opposite nation, can go neutral, and will, if AI.
  Market communists will also go neutral, regardless of relations with USSR, to allow for a thaw with capitalist nations. This will tie into the next PR after this one.

Other than visible changes, the bloc decisions have been rewritten not to depend upon the USA and USSR existing in their normal forms. This could potentially allow, for example, for the continuation of the Cold War even if one of the superpowers forms a superstate. However, this would need to be changed in the "end of the cold war" event in cold_war.txt if this is desired. Either way, it is possible. 
  It could also be possible to allow another great power to take over as leader of a bloc if one gets consumed by civil war or falls lower in the GP ranking. This could be a planned feature.

Please look over the changes with regards to sanity and gameplay, especially AI_will_do logic. These were tested for compatibility, but not for balance or in a normal campaign. If this is merged, I will probably submit my communist changes as-is - will discuss in next PR.


settintotrieste:
Thanks, this is excellent. However there are three problems with changing the country modifiers to triggered modifiers. The first is that it is very difficult to figure out which bloc the country is in during a game, especially when I do tag switching for feature testing. Another problem is that the modifiers are used for other features so changing them will affect the functionality of those. Lastly triggered modifiers tax the performance of the game much more heavily than country modifiers as they are checked and re applied every month rather than just once/twice for country modifiers. So I would recommend using a few triggered modifiers as possible so as to prevent the game from slowing down too much especially in the late game. This is a great addition but unfortunately due to the reasons above, I can only add it in if you revert back to the country modifier system.

ARR8:
Sure. Addressing points - 
1. It is relatively simple to tell which bloc a country is part of by hovering over the prestige, as this will list all ticking prestige modifiers, incl. triggered ones.
2. As far as I can tell, nearly all of the other code uses the flags from those decisions. Either way, anything that uses the modifiers should be rewritten to use the flags or vice versa for consistency. Plus, wouldn't that code still work if the modifiers are there, but triggered?
3. The reason I changed it was to try to save some performance, as the only other way to apply the independent bloc modifier for nations without any bloc membership would be an event that would fire for every nation that has no bloc, including newly-formed nations.

No. 3 is probably the trickiest one here. Originally, I had moved only the independent_bloc modifier into the triggered modifiers, but I had also removed the flags from the decisions and only set the event modifiers, intending to trigger i_b if there were no other bloc modifiers. It seems that triggered modifiers can't check for event modifiers, as the independent_bloc trigger never went away.
  If there's a workaround for this other than using some kind of terrible flag-modifier combination system like existed originally, I would prefer to use it. Otherwise, it's a choice between this and a meaningless event for every neutral country, which IMO is even worse and breaks the flow of the game slightly for the player.

settintotrieste:
1. Yes however it is not intuitive for most players who do not normally use or know that as a means to check bloc alignment. I myself did not know about this until now and I've been playing Vicky for at least 3 years.

2. This bloc system was my very first piece of programming for Victoria so I did not know that country modifiers could be checked hence I used a flag-modifier system but since it worked adequately I felt that it was not worth the effort to change it. Of course ideally it is better to make everything consistent and "optimised" however as I have limited free time, I need to strategise the use of it. It is always better to devote the time and energy doing something new than to redo something that already works with no change to the gameplay. After all, if it's not broken why fix it and then spend more time fixing something that worked originally? However I still very much appreciate your rewrites though I will request that it doesn't try to redo something that already works with the potential of needing more revision of old features.

3. Independent bloc is just a temporary bloc to motivate nations to align to one of the 3 real blocs. It is intended that the country with such a bloc quickly transits to either Demo, Comm, NAM. There is no event that applies the independent bloc to any country and there is no need for one. The only time it is applied is when a country leaves a bloc and is immediately removed when a country joins a bloc.

Going forward, what I will do would be to curate the changes to add in only the main feature of this PR: "The main gameplay change is the expansion of the NAM. Nations which are democratic/communist, but in the sphere of the opposite nation, can go neutral, and will, if AI.
Market communists will also go neutral, regardless of relations with USSR, to allow for a thaw with capitalist nations." So you can still get the ideas that you wanted to implement in the next PR into the mod.

ARR8:
1 and 2 are good points, but I must disagree with 3. Since the NAM only comes after a certain point in the game, a player can simply choose not to join it and not suffer the negative effects, which, to some, may outweigh the positive effects of the modifier. Without the independent bloc being even worse to offset that, players who wish to stay neutral may simply never interact with the bloc system.

Briefly for 2: I agree with this in principle, but the reason I ended up rewriting these decisions is because it was too difficult for me to implement just my changes as I couldn't really understand the code, and rewriting it helped me with that. Vic2 code isn't immediately clear, and variables aren't so obvious in their use as they are in some other environments. I figured I would clean it up while I was doing that. I didn't set out to rewrite it, but I don't feel the time was wasted, especially as it led me to look in some other files that I'll now keep in mind as I work more on this.

Further, I'd like to request that my potential and allow changes be implemented as well, and my changes to some of the effects and other files, like the NAM setup event and localization. Those fix some bugs and rewriting the scopes to use random countries instead of America and Russia shouldn't cause any problems but would allow me to make a future change I am considering.
  Really, everything but the modifiers files and removing the modifiers lines from the effects should be beneficial and noncontroversial. The issue of the independent bloc could be solved with an event at some point, or making it available from the start of the game.

ARR8:
Let me elaborate, in addition to the above, in why I did the changes in the way I did, that doesn't seem necessary at first glance.

I was planning on doing a full rewrite on the NAM as part of my rewrite of Yugoslavia and would allow the NAM to become a real bloc and a third faction to the CW, based on Tito's actual plans. This would be difficult in game, but, if it ever happened, I planned on setting a global flag which would disable the NAM bloc modifier and trigger a different "NAM2" bloc modifier instead, one with only positive effects, for all NAM nations. That is why I couldn't leave the decisions just setting a modifier, because otherwise the only solution would be to add two more decisions to nations to leave the "old" NAM and then join the "new" NAM.

All in all, it's a relatively small feature that hasn't been worth the effort and arguments here, and it looks like I'll end up scrapping it - but now you'll know why.

settintotrieste:
Sure, I will add in he allow and potential changes and thank you for the explanation and clarification, I now understand the reason for the move to a triggered modifier. What I can do to change only the NAM bloc to a triggered modifier but keep the other blocs as country_modifiers so as to allow for the changing of NAM effects via the scenario you've outlined.

ARR8:
Thanks, that would be great. I'm not sure what you want to do with the flags and modifiers so I'll refrain from touching this further, but I would ask that you not forget the changes to the other files. Maybe consider editing the PR then merging it, to avoid merge conflicts?

Thanks again for understanding - I'll try to be more clear about the intent behind changes I make to existing code in the future.

settintotrieste:
Sure.

PR #106, opened 2017-04-16
Unilaterally stop paying WW2 rep early

ARR8:
In the case of a remilitarized far-right Germany or Japan, it would make sense for them to stop paying.


settintotrieste:
This is great!

PR #107, opened 2017-04-16
New modifier icons for democracy and communism

ARR8:
Using the Allies and Comintern faction icons from HoI3. Positive and negative versions for both, applied to some of the obviously democratic or communist modifiers present. Tried to match the Vic aesthetic as much as possible.

Here's the modifiers.dds file now:
modifiers


settintotrieste:
This can actually be expanded to add more organisation specific icons. I will create an issue for this.

PR #117, opened 2017-04-18
More dynamic ahistorical outcomes

ARR8:
Newfoundland, Six-day war, Trst/Trieste

Vary world events slightly, expanded AI logic

### DIFF
--- a/CWE/events/canada.txt
+++ b/CWE/events/canada.txt
@@ -21,26 +21,48 @@ country_event = {
 
 	option = {
 		name = EVT_1002621_A # Newfoundland Responsible government
-		ai_chance = { factor = 0 }
 		prestige = 5
 		release_vassal = NFL
-NFL = { any_owned = { remove_core = CAN } }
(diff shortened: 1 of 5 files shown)


settintotrieste:
This is great! However there is one thing to note when it comes to ai_chance factor and that is all factor values must be less or equal to 1 if you wish to randomise AI choice. An example would be option 1: factor 10 and option 2: factor 90. The AI will still always choose option 1 as it takes factor 10 as 100% and the AI scrolls down the choices and select the first 100% ai_chance.

ARR8:
Uhh...give me a second.

ARR8:
Fixed, plus a few more. That's really good to know now, and not after I started doing more of those.

settintotrieste:
Yes it thought so too myself in the beginning but apparently it's not the case. There's a lot of inconsistency in the Victoria game's programming which I believe is attributed to different programmers who did the various parts of the code. It's quite telling when some commands are written using British spelling and others in American spelling.

Issue #53, opened 2017-03-21
More varied ideologies, esp. for communism


ARR8:
[/u/EmperorPeriwinkle said:](https://www.reddit.com/r/paradoxplaza/comments/60hsio/a_few_things_about_the_cold_war_enhancement/)

I like the Market Communism/Communism split, Every country, especially the USSR should get a market communist party.
  I'd like ideologies activated over time, actually. I'd like if every major era had a set of ideologies unlocked. Market Communism should unlock in the 1980s, for example.

[/u/bfmv24 said:](https://www.reddit.com/r/paradoxplaza/comments/60hsio/a_few_things_about_the_cold_war_enhancement/df6sels/)

Relating to this, I think it should be included from the beginning because this was a huge sticking point between Tito in Yugoslavia and Stalin in the USSR.

settintotrieste:
Adding a market communism party will be difficult as I have to manually add the parties to every country file so I will just add the party to the Soviet satellite states for now. I will leave market communism avaliable from the beginning.

ARR8:
Concerning the second point, here is a map of all the ideologies currently in the mod:

| | A<br />Always support political reforms | B<br />Support political reforms with militancy | C<br />Support political reforms but revoke with militancy | D<br />Support or revoke political reforms| E<br />Revoke political reforms but support with militancy | F<br />Revoke political reforms with militancy | G<br />Always revoke political reforms | H<br />No political change
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| **1 Always support social reforms** | Progressive | Socialist | | | | | Communist<br /><br />Market communist
| **2 Support social reforms with militancy** | Liberal | | | | Populist
| **3 Support social reforms but revoke with militancy** |
| **4 Support or revoke social reforms** | | | | Moderates w/ militancy | | | | Nationalists if in power
| **5 Revoke social reforms but support with militancy** | | Conservative
| **6 Revoke social reforms with militancy** | |
| **7 Always revoke social reforms** | | | | | | | Traditionalist
| **8 No social change** |

To keep it symmetrical, there is room for ideologies in A6 and D8 to counter communists and nationalists, respectively. Some possible ideologies that are not strictly represented are pirate parties (possibly considered progressive), libertarians (possibly liberal), anarchists (relevance debatable), or single-issue parties (difficult to implement).

[...]

PR #120, opened 2017-04-22
Communist party conference mechanic #53

ARR8:
This uses pseudo-election events to signify a party conference for a possible transition from communism to market communism or democratic socialism. The party will have a choice in each event between keeping communism or reforming slightly for an economic or political benefit. They are not true election events, only skinned as such, because they represent an actual choice to be made, but the party selection does happen by election of the upper-class pops, so loyalty national foci and such will actually have a purpose for communists.

This is supposed to be a streamlined and generic system; it only adds one flag, and uses existing modifiers/pictures. The plan is to modify/add each historical reform event chain, such as Tito or Deng Xiaoping, to also change support for different kinds of communism depending on whether the reforms were made, and then run this event chain at the end. This way, those countries will likely adopt market communism as a party, but a player can choose to steer it in a different direction. The design philosophy for those will probably be similar to these generic events, with a choice between a bonus or keeping communism, but using historical changes in those nations. The events added here are also based on real events that happened, but made generic.
  I anticipate tying this system into historical events, such as Hungarian and Polish revolutions, and adding a couple of alternative history scenarios for more distant Soviet satellite states. There will probably be a trickle of these kinds of events until the 90s, when there should be many of them.

It is possible to transition to a socialist democracy from this, but it is very difficult. I'm not sure if socialist democracies are actually possible in any other way in the mod as-is, but this is one way to reach them.

Using the bloc mechanics, this will also create a large number of states transitioning away from the eastern bloc around the 1990s, and has the possibility to trigger the fall of the Soviet Union if enough states leave. Combating this is very difficult for the USSR, and would require military operations and careful sphere management. There is a possibility for a renewed NAM bloc as discussed.

Something else added is an event for the USA similar to the Iran-Contra affair, in event no. 8023334

This is **not** finished! It will not work unless every country this can possibly run for has a market communist party. I'll work on adding those. It is also untested in a real game, so possibly ai logic needs to be tweaked for the election events and for willingness to fight breakaway states. I've disabled some events from running while the conference is happening, but I think there are others that need to be changed, as well. Additionally, there may be an issue with the proletarian_dictatorship1 government, but I'm not completely sure what that is so unsure how to account for it.

Please look it over! This will probably need some rebalancing in addition to the changes mentioned above. Also let me know if you have any problems with the implementation.


settintotrieste:
Great work ARR8, this is tremendous! 

Let me address some of the issues you raised, the proletarian_dictatorship1 is the same as a normal proletarian dictatorship as it was created to allow for Vietnam to be named North Vietnam so you add it together with the proletarian_dictatorship.

If I remembered correctly there is a "transition from communism to democracy" decisions file with a decision to turn to socialist democracy, a Chinese tiennamen event chain that does lead to a socialist democracy and the success of the August coup chain also leads to a socialist democracy event so you may want to give them a check to ensure that there are no clashes.

All in all this is still excellent, so what I will do is to review it by the next weekend, which should give you time to do further checking/tuning if required and to add the necessary market communist parties before adding the feature in.

ARR8:
Sounds good!

settintotrieste:
May I know if you've managed to add the necessary  market communist parties yet?

ARR8:
Yes, I can commit that right now. Every country has an identical m_c party with a generic name.
  I'm trying to figure out how to name them, though. Trying to use regular expressions to extract the tag, now reconsidering the whole ordeal.

ARR8:
I tried that at first, but it didn't work well in testing, while the current implementation seems to work alright. Maybe I did it wrong; can you give an example of what you mean?

settintotrieste:
Sorry, I wasn't very clear in the first post. What I meant is that the events with trigger "has_country_flag = communist_plebiscite" and "election = yes" all have an mtth of 1. This means that all of those events will fire at the end of the first month after the election begins. I believe it would be better if those events had different mtth so they won't all fire by the second month of the 6 month long election.

ARR8:
In testing, it seems like they fire randomly enough. Wouldn't a different MTTH only affect their rarity? I'd prefer if all of them had the same MTTH, though if there are too many events it can be raised to two months or something.

settintotrieste:
Alright, it's good to hear that it works well in practice. I will merge the PR now.

ARR8:
Great - next priority will probably be reviewing historical event chains as mentioned and merging all transition from communism instances to use this system. I'll probably schedule most of those for 1.07 to make sure these work fine.

settintotrieste:
Sure

Issue #116, opened 2017-04-17
Optimizations - POPS


EmperorRAG:
This thread is for the continuation of the discussion relating to pop optimization that was started in #109:

Initial post

> I was taking a look at how Victoria 2 does calculations involving pops and I saw that the mods performance could be improved. If you were to make craftsmen and capitalists stay only in the state capital, you would reduce pop fracturing. The reduction of pop fractures would decrease the amount of calculations required for pops in the game session which would increase the game's performance. The proposed change would also tie into the urbanization patterns that follow industrialization therefore making the mod more realistic. Additionally you can go further and make it so that all pop types besides farmers, teachers, laborers, entrepreneurs and aristocrats only present in the state capital. The results of the additional change should greatly increase performance as well as aid the game in performing pop promotion and demotion. The most noticeable impact from the changes will most likely be reduced RAM usage, although it would be dependent on the user's PC. 

> 

Pop logic

>  the pop promotion to craftsmen isn't limited to state capitals, pops in all provinces will continue to promote to craftsmen but on promotion will teleport to the capital. 

> 

> All that needs to be done is the changing of pop types; so an example could be 5000 craftsmen in a non-state capital province being changed to farmers/laborers and then in the state capital province changing 5000 farmers/laborers to craftsmen.

> 

Solution to the crash

> The crash comes from the pop type needing to be in a state capital province but due to it not being in the state capital province the game crashes, this issue is caused from the pop types not being teleported to the state capital province as they weren't promoted or demoted.

>

ARR8:
From the other discussion:
  state_capital_only is a built-in property of pops, which in vanilla is applied to aristocrats, capitalists, and bureaucrats, and in the mod currently only for soldiers and officers.

The migration factors already include some urbanization logic:
>`	modifier = { factor = -1 location = { is_state_capital = yes } }`
>`	modifier = { factor = -1.0 OR = { type = farmers type = labourers } }`

If, for any reason, we can't move all the pops into state capitals, we can instead strengthen this heavily for the desired pop types, so that they will probably migrate to some state capital on their own. If needed, we can also add a push for farmers and labourers in state capitals or those with urban terrain for a "ruralization" though I imagine this isn't as necessary.
  This would still lower calculations by merging pops, but it is unnecessary and not preferable if the former is possible.

EmperorRAG:
[...]

PR #124, opened 2017-04-22
Optimizations - POPS #116

EmperorRAG:


settintotrieste:
This is excellent, though I need to know if it would affect split states pop promotion. If it doesn't then I will add it right away.

EmperorRAG:
It doesn't affect split state POPS as split states have a capital each. Upon unification, the state capital is combined to the one defined in the games files and the pops in the previous capital are transported to the new capital.

settintotrieste:
Thats great news, I will add it in.

Issue #108, opened 2017-04-16
Modifier icons + UI rework


settintotrieste:
New icon pictures for organisations/nuclear war etc can be added as country modifier icons to make it easier to differentiate the modifiers and increase aesthetic appeal of the politics screen.

ARR8:
Step-by-step process of how to make the icons as in #107:
1. Expand modifiers.dds by 32 pixels horizontally for every icon to be added.
2. Put the icon centered in its 32x32 block.

If tinting icons red and green:

3. Desaturate based on luminosity.
4. Colorize. Generally, if using HSL, the hue for green is about 98 and the hue for red is about 359, saturation should be unchanged at 50, and brightness adjusted to match neighboring icons if needed.

Whether tinting or not:

5. Edit interface/general_gfx.gfx line 17 to reflect the new number of icons.

When testing, keep in mind [the numbering offset](http://www.victoria2wiki.com/Modifier_effects#Modifier_Icons).

ARR8:
Would a nuclear icon be helpful? ![nuke_icon](https://cloud.githubusercontent.com/assets/8064135/25307175/5a2d26dc-276a-11e7-9a4c-922304c7244b.png)

ARR8:
I think this would also be the issue to discuss the aesthetic of the mod. NWO has obviously departed from the Vic2 aesthetic, which is good. However, I think we can agree it went too far in some respects, such as the photorealistic map changes, which this mod undoes.

Most of the graphics are fine, but there are some which I feel can be improved. Some examples (mod vs vanilla):
- Simplified Muslim icons
  - ![icon_religion](https://cloud.githubusercontent.com/assets/8064135/25307303/c660814e-276c-11e7-9e19-49fc4f64f8a7.png)
  - ![icon_religion 2](https://cloud.githubusercontent.com/assets/8064135/25307304/c7f28836-276c-11e7-995b-15771cf3f1a3.png)
- Influence investment stars
  - ![diplo_priobutton](https://cloud.githubusercontent.com/assets/8064135/25307327/3253d248-276d-11e7-864d-afb1f8e6fb5a.png)
  - ![diplo_priobutton 2](https://cloud.githubusercontent.com/assets/8064135/25307334/52af0d1e-276d-11e7-924c-809646fecc6d.png)
- Grayscale tech boxes
[...]

PR #126, opened 2017-04-24
Icon changes #108

ARR8:
The changes mentioned in #108.

Religion icons have been changed further. For reference, here's the religious icons now, though some are unused:

new icon_religion

Other than reskinning, there were some changes made to religious icons:
- Most notably, secularism now uses the Hellenic pagan icon, a stylized Parthenon, from CK2 rather than the shamanist icon from EU. I feel this is an improvement since the shamanist icon would be familiar to anyone who played EU while the Hellenic icon is unused except for creating custom rulers in CK. It also seems more appropriate to me, as it could actually represent "Western secularism."
- Judaism changed to the menorah. No big reason for this, but the menorah icon seemed to fit in better with the other icons aesthetically. The old one is left in for any reversion.
- Ibadi given its own icon, so no longer identical to Sunni. The icon is from CK2. The other option and the one used in HPM is the EU icon, a blue-tinted star and crescent, seemed lazy to me as a palette swap, even though it might be more familiar.

Buddhism now has the vanilla 3 icons to choose from. The last four icons are unused: Druze, Mormon, Yazidi, Jain. The fifth is also unused: coptic.

Something I'm not sure about is icon_religion_small. They are defined in the interface files, but seem to be unused. They do not match their corresponding larger icons in vanilla, and its vanilla contents imply its a holdover from EU3. I've taken it out for now, and it doesn't seem to cause any problems (falling back to the vanilla file).

National values have been returned to vanilla, but I did have to make one for Islam. I tried to match the vanilla style:
v2_6

One tangential change is a bugfix where desert and desert hills terrain looked the same; desert terrain changed to match vanilla.

Some screenshots:

v2_1
  The diplo-sliders are a little bright compared to the background, so those may need to be dimmed.

v2_3

v2_5


settintotrieste:
Excellent work!

PR #128, opened 2017-04-25
Israel fall premature fix

ARR8:
Prevent fall if owns Sinai and Camp David not yet proposed, MTTH tweaks. As discussed in de1b76e

### DIFF
--- a/CWE/events/israel.txt
+++ b/CWE/events/israel.txt
@@ -9,7 +9,7 @@ country_event = {
   picture = "reparations_agreement"
   is_triggered_only = yes # GER
   option = {
-    name = EVT_8007721_A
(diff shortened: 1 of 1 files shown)


settintotrieste:
You might want to include not own the Suez Canal Zone and the Golan Heights as well as they do contribute a lot of Arab pops as well which can make pop_majority arab

ARR8:
Well, I would want Golan to contribute to the fall of Israel since there's currently no way to give it back. SCZ is part of the Sinai. I only include the Sinai since it's generally only occupied for a short time.

settintotrieste:
Alright, thanks for the clarification.

PR #131, opened 2017-04-27
Use width of politics window for modifer icons

ARR8:
Start near the top right and expand left. Works on lower resolutions. Screenshots:

Many icons
  v2_7

Few icons
  v2_8

I haven't found any problems with this approach, though I have only tested it briefly.

It's possible that it looks cramped. I'd value an opinion on this; maybe it would be better to align the left edge with the vertical divider rather than the edge of the window. Either way, I think it's an improvement on the current system, and does let you actually browse your modifiers.


ARR8:
The option mentioned above:
  v2_9

settintotrieste:
This more compact layout is better. Please add it.

Issue #149, opened 2017-05-02
French colonies in India


settintotrieste:
Now that we have added the smaller Portuguese colonies in India, perhaps we should also add the French smaller colonies in India as well for consistency? Karikal and Yanaon on the Coromandel Coast, Mahé on the Malabar Coast and Chandernagor in Bengal can be added as well in a similar fashion as the Portuguese Indian colonies by making them enclaves of Pondicherry. This is historically accurate as well as these small colonies were under an administrative union with Pondicherry.

ARR8:
Gladly, but I would like to know what you would like me to do with their populations. For the Portuguese colonies, I added them to Goa, but I wasn't sure if I should have reduced any from the provinces I took pixels off of. I'm not too worried about it since India is slightly underpopulated now after #133, but it'd still be nice to know.

settintotrieste:
Sure, you can do the same as with Goa and add the population to Pondicherry without removing pops from the affected provinces. The under population of India should compensate for the extra pops from the French colonies.

settintotrieste:
![invasion-of-portugal-1024x933](https://cloud.githubusercontent.com/assets/24409479/25658217/5b760cc0-3034-11e7-90b1-9a8a0ca81e05.gif)

Based on this map, only the colonies of Diu, Karikal, Goa and Pondicherry are really visible on the map so I think we should correct the current setup to only display these colonies on the map.

ARR8:
I'm of the opinion we should add every coastal colony to allow naval docking in case we ever change the sea tiles

settintotrieste:
This is something that we can do though I'll prefer to keep this on hold first. We can see revisit this again for 1.07.

PR #153, opened 2017-05-03
Size for Indian colonial ports

ARR8:
The new exclaves are very undersized if using Goa and Pondicherry themselves, as well as other provinces like Gibraltar, Venice, Gaza, Hong Kong, and Macao as examples, even though it may be more realistic. Generally Paradox games will do this all the time for usability.
  Even though they're not provinces, I would still highly suggest the new exclaves be at least the size of the cursor so as not to frustrate a player who doesn't know they're not provinces at first, with size on the map projection a secondary concern.

This PR simply restores the size of the Portuguese ports and expands the new port of Karaikal.

I think that also removes some of the reason not to include the smaller ones, which would be good to include even if just to lift fog of war in more coastal provinces where the colonial empire has a port.
  However, pursuant to #149, I'll hold off on those for the time being.


settintotrieste:
I think that it is better not to increase the size of the colonies if they are not actually provinces as it would confuse players. The small size of the colonies is in a way a means to indicate that they are not really provinces.

What I propose that we should do now is to turn the colonies into full provinces which makes it less confusing to a player that does not know that they should be connected to Goa. This can also add an extra layer of strategic gameplay as ports can be added to those provinces and let the decolonisation of India have more outcomes like France losing Pondicherry but keeping karikal. We can make those provinces the size Gaza to be smaller than Pondicherry and Goa but big enough to be clicked.

ARR8:
I agree that those outcomes could definitely be done in the future, but for now large exclaves of existing provinces exist in vanilla. I can't think of any off the top of my head, and I think most of them are island chains, but enough areas are enlarged that I would advise against departing from it now

settintotrieste:
Alright, I will see what I can do in the meantime about the colonies.

settintotrieste:
I've decided that I will scrap the idea of adding the Portuguese and French colonies in India as both choices of making the colonies full provinces and leaving them as it is are unsatisfactory.

ARR8:
I'm very disappointed to hear that. May I ask what you dislike about the current system?

settintotrieste:
I didn't want to abandon it either but after playing a few test games, I felt that it was jarring for there to have the same province spread out like lslands on the map of India visually in addition to the confusion for players who don't know the reason why Goa and Pondicherry has an enclave. Players would have a high chance of thinking that the representation of the colonies is a bug. 

Moving forwards, what I am planning to do is to change the representation from on the map to province modifiers similar to how Microstates were done in the game.

PR #154, opened 2017-05-03
Seculiarization and desecularization

ARR8:
As discussed in cc44072, this is (atm) a proposal to fix the religion selection in desecularization to always select the correct national religion.

The way this works is by selecting a random province of the primary culture, if there are any, and checking whether any pop has a certain religion. Then, it checks whether the capital also has a primary culture pop of the same religion. If both check out, that is made the official religion. This should almost always select the correct religion, even with crazy immigration and assimilation. It should only not work if there are only secular pops or no pops of the primary culture. This may be possible with extreme secularization and a small starting culture, such as in Yugoslavia, but then it would make sense that no one remembers the old religion.

This is fully dynamic and history independent - if a resurgent Poland conquers a Belarusian state, convert every pop to Catholic, then release Belarus, it will become Catholic by this method, even if every other Belarusian in the world is Orthodox.

There is a possibility of, for example, Germany running this and selecting a province in Bavaria that happens to have absolutely no Protestant pops, while Berlin has at least one Catholic pop. In this instance, Germany will become Catholic. However, since this will be redone at the next secular/moralist cycle, I wouldn't be too worried. I suppose it's technically possible for a Catholic party to win one election. (EDIT: That statement would only make sense once Germany is back to being majority Protestant as it should be)

**Please don't merge this!** Look over the effects when you get a moment and let me know what you think; I would like to still implement all the changes we discussed.

Something I've been considering, instead of both, is removing the state religion reform completely, and running this effect as an event whenever a moralist party wins an election/gets put in power, and setting national religion back to secular whenever a secular/atheist party is in power.
  This doesn't have to happen immediately after the election, so that traditionalist and atheist parties will get a MTTH boost. Then, would probably change both atheist and moralist to get a boost to conversion while secular discourages it except for very educated pops.

This should streamline the current system which uses both reform and party policy to a confusing degree, with national religion unclear, and have the plus side of using the game engine more. Thoughts?

edit: Seems I left some comments in the first five capital checks. Just pretend those lines aren't commented.


settintotrieste:
I've given it a quick look through and it's excellent! Unfortunately I am against removing reforms for the sake of maintaining game stability. But I am wholly in favour of your idea to change the national religion to secularism when an atheist party resizes power. I will create the decision system to allow for that flip flopping on national religions according to the religious policy of the ruling party.

ARR8:
No need for the decisions - I was going to create a pair of events with variable MTTH to simulate the amount of time passing until the party establishes a religion/officially secularizes the country. I was thinking around a year into the administration but reduced in half for more extreme ones?

Luckily, I don't think these particular reforms are used very much in game files. I'll give them a search and see if there are any that can't be replaced simply with party policy.

settintotrieste:
Sure, one year is a suitable amount of time with it being maybe around 3-4 months if it is a traditionalist ruling party and 1 month if it's a theocracy?

Great news, do let me know of your findings.

settintotrieste:
That's great! I never understood the original purpose of the religious rights reform before anyway so it's good that it is removed. Now all we need to do is to complete the events to complete the mechanism and I can add in the PR.

ARR8:
Done, with a minor consciousness effect included. At some point, I'd like to go through and make sure countries' state religions match their 1946 starting party, but that can wait just to save one event. I will do a couple of major ones now, though.

settintotrieste:
Good job! This is great! I'll add in it.

ARR8:
Thanks! If you find pops are converting at the wrong rate in your test games feel free to mess with the base assimilation chance in poptypes. I've lowered it from 0.3 to 0.5, but I think vanilla is 0.1

settintotrieste:
Sure

PR #159, opened 2017-05-06
Size reduction

ARR8:
Removed interface files identical to vanilla. Confirmed that nothing has changed in-game.

If any of these need to be changed in the future, they can simply be copied from vanilla first.


settintotrieste:
Great!

Issue #41, opened 2017-03-20
Middle East Civil War fixes and enhancements


ARR8:
If ISIS wins the Algerian civil war (event 8000318, algeria.txt), there is no event to make it a theocracy. This results in strange consequences, especially if Algeria's government form has changed along the way.
![v2_1](https://cloud.githubusercontent.com/assets/8064135/24087869/87b0f85c-0cfa-11e7-9dd2-9a6dcac7e817.png)
  It is able to join numerous international communities, including the UN, and holds elections for the caliph.

As a more general problem, if ISIS exists at the beginning of the Syrian civil war (event 8014730, syria.txt), and Syria wins, the annex war goal leads to things like this happening:
![v2_2](https://cloud.githubusercontent.com/assets/8064135/24087902/dec4211e-0cfa-11e7-88d9-950ac9a3bc7b.png)
  I doubt even the pan-Arab republic would be happy with this one.

It should be a simple fix, just changing the war goal for Syria. It can either be acquire all cores, which will leave ISIS with whatever land they had, or something along the lines of a dismantle, as it makes sense that a surrender from ISIS would involve leaving all occupied land.

On that note, it is currently impossible for a great power to intervene against ISIS due to it being a Civil War war goal. If we want to keep it that way, we could add "call_ally" disabled in the war. Although, I'm not sure it makes sense to prevent a GP from intervening, given all the intervening they're currently doing.
  It could also be considered to simply give the GPs CBs against ISIS so they'll declare their own wars. This makes sense, as our modern great powers aren't exactly cooperating in Syria.

settintotrieste:
Thanks for the bug report! This will need to be fixed. Ideally for now I prefer to keep to using the restore_order CB against ISIS. The problem with using a special CB for GPs is that they may end up stuck in perpetual wars when they declare war against a land locked ISIS in a nation that did not grant the GP military access. This can cause performance issues and perpetual mobilisation in those GPs. 

A compromise solution would be to give states in spheres of GPs and are neighbouring ISIS the intervention CB which is probably more realistic as well as currently the Syrian intervention is still mostly a proxy one.

settintotrieste:
Syrian Civil War has been changed, Yemeni & Iraqi Civil War added as of 3cf26b0. Testing needs to be conducted on those events though.

ARR8:
In addition to testing, the following the planned enhancements, copied from #80:

[...]

PR #160, opened 2017-05-06
Fix Algerian and generic civil wars #41

ARR8:
\+ strings

This fixes the initial problem reported in #41, so that ISIS will act properly and not as a separatist republic.

For the generic dictatorship civil wars, general cleanup. Primarily, the coup outcome event will only fire if the other events came first, so not for just any revolt.


settintotrieste:
Thank you for your work on this. Unfortunately I will have to reject the changes to the generic civil war as I need it to work this way as a means to stop endless wars. I am alright with the Algerian and Isis civil war changes though. I will close this pull request until the changes have been corrected.

ARR8:
I would disagree about using this solution to stop revolts - instead, we should make it simpler for revolts to succeed and end on their own. A simple fix for this would be successful rebellions if they control a certain percent of the country, rather than just the capital. Also, something HPM does is apply a modifier chain if provinces are constantly occupied eventually leading to independence for that area if overseas, to prevent constantly occupied territories.

This solution is also not ideal as it assumes that the rebels are military-backed or right-wing.

Do you have any examples where this is a problem?

ARR8:
By the way - that generic system does break the Algerian civil war, which is why the changes were needed. I would still prefer better rebel logic to adding an exception for Algeria, though. As it is, I can't include the Algerian and ISIS changes if the generic CW system isn't changed.

settintotrieste:
At this moment I prefer if it wasn't changed. I can make an exception in the generic system to prevent it breaking the Algerian civil war.

ARR8:
I'd rather not do that for every single civil war, as I'm planning on doing several of them. Wouldn't it be a better solution to encourage rebels to enforce demands on their own? After all, if a country has even a few states occupied nowadays it usually capitulates.

In addition, we've done enough tweaking to mobilizations and such that I doubt drawn-out revolts are really still a problem anymore. If they are, I haven't seen any that wouldn't be easily fixed. Can you give some examples that I might be able to test?

settintotrieste:
I understand your point and my apologies for not explaining my position well. The reason why I do not wish to alter the generic events is because it is a matter of principle for me that generic events should not be tailored for scripted events as they often they create a lot of gameplay flaws. It is much easier and safer to solve the problem by creating an exception that only affects one scripted event than modify the function of the generic events which can affect everything. Furthermore, rebels can at this point enforce demands on their own but I prefer not to rely on them completely as they are very unreliable.

ARR8:
I see, but really the generic system does not work as well as a system using only the game mechanics could. For example, say the Algerian CW used only the generic system, the historical Islamist rebels would find themselves fighting for a liberal democracy or the military could have a 5% chance of taking over and fulfilling the actual goal of the civil war, a theocracy.

Additionally, the system entirely breaks separatist rebels and prevents them from ever attaining independence. Frankly, all of the outcomes of the generic system (ignoring the event chain, which is actually very good) could be emulated using the game rebel system. Even if it is unreliable, it's probably the only way I can think of to tie into movements and radicalism. The outcomes in the generic system lower militancy across the whole country but, in my testing, this tends to only delay another rebellion because the movements never disappear.

It would also allow preventing a rebel victory without needing events to cover every possibility. We already have UN peacekeeping missions, and I'm sure another country could simply march in and deal with the rebellion in any case, which the system doesn't account for. It doesn't make sense thematically for a rebellion to enforce demands if the Soviet army is gunning down the capitol, and it gets even more complicated when hostile armies are taken into account. This happened in my game, and was frustrating.

I would urge you to reconsider simply easing rebel enforcement to raise reliability and tying into vanilla systems whenever possible, because modders simply don't have all the tools to deal with the very complicated game rebellion system. While we should create as many event chains as possible to initiate rebellions, like the generic CW event chain itself, I can't think of a way to handle the actual rebellions as well as the game rebel system or a two-faction war, which would be a pain to implement for every single CW.

settintotrieste:
I've done a test of the Algerian civil war event chain and I didn't see any problems arising from the generic civil war chain.

settintotrieste:
It is technically not true as the final stage of the event is to cause all provinces with militancy above 10 to come under rebel control allowing for the separatist states to secede when all their cores are under rebel control.

Such a scenario has occurred before in ww1 Russia. When the war was going badly against the Russians, the Russian revolution began which also involved foreign forces on Russian soil.

The current generic chain is extremely simple to avoid for scripted events. Just do not create an event that makes militancy rise to 7 and above militancy and if you wish to trigger an unscripted revolution, just make the event add militancy to above 7 and the civil war mechanics will handle the rest. 

The vanilla system works alongside the generic system. The vanilla system works for separatism as they often only exist in a small part of the country to prevent the national average militancy to be above 7. 

At the end of the day I do not find that there is a need to shift it to the vanilla system when the generic chain works alongside the vanilla system.

ARR8:
I can send my save file if you like. There, it is impossible to lose the Algerian CW because the generic event fires first. It is a very common scenario for militancy to rise on its own. Keep in mind the Algerian CW is a special case, because it fires the first time the country is not a republic. At that point, militancy will probably be very high if it happens after a long time of republican rule.

Ultimately, I can't simply prevent event civil wars from rising high in militancy because I plan (or did plan) to add prelude events to raise militancy in the background conditions of the war. Simply put, if you insist on this, you won't find me touching civil wars. I can't account for finely tweaking militancy levels to prevent this event from firing and ending the war early on an outcome that would never happen with that specific civil war.

Forgive me, but I'm not sure what you mean about generic event chains with scripted events. That seems to me exactly what you're doing - the generic outcome fires for even scripted war chains. If you mean the vanilla rebel system is the generic event chain here, I will defend using that simply because it allows players to prevent the rebellion beyond hoping to get the right dice roll on the percent chance of an event, and allows for interaction between the rebels and multiple countries, such as occurs in nearly every modern rebellion.

In the generic system, it is impossible for separatist rebels to enforce demands by reaching the capital, because militancy has to be so high for them to appear that the event is likely to fire first.

If the Russian Revolution were to happen in the mod, the Reds would never win because the generic event would fire and make the head of the Red Army a White sympathizer who installs a nationalist dictatorship before the rebels can enforce demands naturally. This is my main problem - that makes sense for generic civil wars we don't know all about, but nearly every scripted CW is a special case.

I really can't understand what you dislike so much about the vanilla system. It is possible to tweak pop loyalties to tailor which rebels appear, and having simultaneous rebellions seems realistic to me. I'm sure anything else wrong with the vanilla system, such as weak enforcement, can be easily modded.

settintotrieste:
There is a simple solution around this. I will add a check for a "specially_disabled" flag which presence prevents the events from firing. The flag can then be applied at the start of the chain you wish to create and then removed at the end of the chain. This will circumvent the potential risk of this interfering with the events.

I am slowly separating the generic events from the scripted ones so as to allow for more reliable outcomes.

The problem with the generic system is that changes in government these days really don't tend to be massive and protracted uprisings that involve 10s of 1000s of rebels slinging it out across the country. And the vanilla system without the generic chain is really the only way to change the government type. The generic chain simulates the more contemporary type of revolution. A quick short insurgency that often ends in swift coup followed by further insurgency. Of course it is not easy to control the number of rebels that spawn but at least there is more of a control over how long the rebellions last. 

Separatist rebellions are generally the ones that are like the vanilla rebellions and they are not affected by the events as separatist rebels do not need to capture the capital to enforce their demands. They just need to conquer all their cores.

settintotrieste:
I have re-looked through your alterations to the Civil War chain and I realised that I have misunderstood the changes. I have now added in the key change which is coup outcome event will only fire if the other events came first, so not for just any revolt. I apologise deeply for starting a long argument over the misundertanding. I will look more closely at the changes in future.

ARR8:
In that case, I've reopened the PR to also include the flow and string changes I've made, since those seem to be uncontroversial.

I don't mind the misunderstanding, but please, in the future, don't merge just a subset of the changes from a PR if we're still discussing it. That really rubbed me the wrong way and felt like an end to the discussion.

settintotrieste:
Sure, I shall try not to in the future however I don't want the other changes to the chain as I want to preserve the flavour from the change of government events and I need to keep the modifiers after the end of the civil war to simulate the instability of the post civil war stage and prevent another one from appearing too quickly. As for the localisation, I am alright with it except the one renaming government house to Capitol as it sounds rather specific to America.

ARR8:
Alright - make sure you don't forget the OR clause at the beginning of the trigger for the outcome event, as this is what prevents it for other wars.

settintotrieste:
Sure, I will add it in as soon as I get back to my computer.

PR #162, opened 2017-05-07
More permanent Hungary fix

ARR8:
Found the bug causing the Hungary problem - the event that fires on a puppet war victory was made before there were multiple communist ideologies. It's been adjusted, and, just in case, the two events to set Hungary back to Soviet have been merged.

Also tweaked market communism to allow at war, but only for vassals. Related to Hungary tangentially.


settintotrieste:
Great!

PR #167, opened 2017-05-10
More communist bugfixes

ARR8:
This should fix the repeated war bug from both Hungary and communist Austria. I've made the fix from last time more lenient in application. Additionally, found a new bug with the Sovietization events - they do not change ideology to communist, thus leading to a problem specifically with Hungary and non-communist Austria, which both start with socialist parties.

Hopefully, that's it. I've removed the event again, and am hoping we won't need it anymore.

Fixed another bug where said Austria would join then immediately leave the eastern bloc

Enabled the ultimate two events in communist_plebiscite from a bad scope


settintotrieste:
Thanks

Issue #19, opened 2017-03-17
US-Soviet Joint Projects


settintotrieste:
Main feature for 1.06

ARR8:
Some possibilities:
- Joining the ISS project could give relations with other members, easy change
- After a certain point, and with low tensions, the bloc leaders could let members of their aid systems (Marshall plan, COMECON, also possibly international) apply to aid from both powers - joint economic development, investment in the third world. 
- If both have similar attitudes to climate change, joint environmental projects or experiments, especially to do with empty land in Siberia and Western America
- Events to do with Olympics if both are involved and neither caught doping, visits from teams, games between the powers
- Events between the capitalist bloc leader and any market communist state to foster relations and trade (incl. China)

settintotrieste:
Those are excellent ideas! I will try to get as many of them in as possible.

ARR8:
Great! Just submitted a quick patch for the first one, and I'll look out for other suggestions.

settintotrieste:
Thanks!

ARR8:
Other ideas:
- If the treaties after SDI development are completed, or at least just SALT, the two could work on global disarmament. I doubt any power would agree to remove nuclear weapons completely, but I think some of the arms race items could begin to be rolled back worldwide.
- If the USSR begins Glasnost' reforms, there could possibly begin a student exchange between the countries, something that could encourage liberalism and democracy in the USSR and sympathy for socialism (once #50 is fixed and it becomes less powerful) in the US.
- Beyond the interpol, there could be counterterrorism efforts: the USSR historically put in a lot of effort suppressing Islam within the country. Possibly one of the ways to end the war on terror could be the adoption of the Soviet tactic worldwide by both blocs of pushing a moderate, decentralized Islam to weaken radical movements. However, there were certainly large differences between the powers on their ideas about religion, so this should probably be very difficult.

settintotrieste:
Nuclear disarmament that reduced the size of the arsenal which was what the superpower this would be difficult to model into the game because currently the nuclear system is based off delivery capabilities instead of arsenal size so I probably wouldn't want to add that in. Currently nuclear acquisition is limited by the nonproliferation treaty which most nations will sign.

This is a good idea. I will put it under a "US-Soviet cultural outreach" decision.

[...]

PR #180, opened 2017-05-16
Joint Projects rewrite #19

ARR8:
Lot of changes here - most relocalization and new events will be a future PR.

I've renamed all of it to Cross-Bloc Detente so it would show up higher in the decisions list.
  The superpowers will also drift closer together in ideology if they enact certain projects, using market communism or socialism as the common ground.
  Some of the more controversial changes open the communist bloc up to schisms, using the communist conferences.

And a lot of other changes. Notably, everything should now be more dynamic and tie into tension levels.

All in all, I think these are a great addition for 1.06! Maybe one possible end to the cold war would be a closer ideological merge between the powers once maximal detente and integration is reached? Either way, that's probably a discussion for a separate time.

Something I wanted to ask: I notice all of the detente decision can technically be picked in any order, though they do depend on technology. Do you want any of them to be prereqs for others? E.g. open borders almost seems like the "next level" of open market, and ditto with research and space. I've kept it the way it is, but that can definitely be done for future joint projects.

EDIT: Also, all untested. Validator had no complaints, though.


settintotrieste:
Excellent work, this is great!

Yes, it will be better to use previous decisions as prerequisites in tandem with the technology requirements.

I wholly agree, that is a good possible way to end the Cold War. I'm thinking of tying it into the existing USSR reforms decisions. Perhaps a period of detent via the joint projects first before the start of the reforms to end the USSR? This will probably require tuning of the start of the joint projects to often happen during the 1970s.

PR #181, opened 2017-05-17
More Joint Project fixes, prestige #19

ARR8:
Fixes bugs:

- The Far Eastern Republic can no longer renounce world revolution
- Player can't abuse the system by taking a joint project decision while an event from another one is up
- Continued potential and flags cleanup
- End of the Cold War event now removes joint projects, cleanup

And the winner of the Cold War will now get prestige based on how many detente efforts were made, up to a maximum (currently) of 140.
  **This actually doesn't seem to work, for some reason. The code looks fine to me, maybe you could take a look at it?**


settintotrieste:
Thanks!

The problem lies in the checking of non whole number variables. Variables must be in whole numbers.

ARR8:
I should probably mention, I tried that, too, but it still wouldn't work with the var equal to 1. Maybe they all have to be decremented by 1?

settintotrieste:
Perhaps, you can give it a try

PR #185, opened 2017-05-19
Ideology and national focus rework

ARR8:
Add centrist and market communist, rebalance loyalty values*, parliamentary coalitions, redo icons #108

\*I think the changes from b944216 were applied to the wrong set of foci, the already-weak NGOs. I've changed the internal loyalty foci instead to a logarithmic midpoint between what was present and vanilla as user feedback suggested. This seems to be more balanced.

For coalitions, I've gotten rid of the minor group and merged populists in with the right and nationalists in with the center. I wasn't sure where to put them, especially seeing as some nationalist parties, like the SNP, are left-leaning, and some, like the Rwandan leadership, were right-leaning. I've renamed them, respectively, to "Right and Far-right" and "Centrist and Single-issue."


settintotrieste:
I will need to reject this PR as it tampers with the ideologies. Please do not touch the ideologies, this is a core feature of CWE that I do not wish to change.

ARR8:
Please check again - ideologies are unchanged, just the coalitions

settintotrieste:
Sorry I meant the coalitions, I am fine with the national focus changes

ARR8:
Are the coalitions needed for anything? They don't really make sense right now, with progressives grouped with moderates and etc.

settintotrieste:
Yes they are there to decide the elections in the case where no single party has a majority.

ARR8:
I'll revert those when I get the chance. How about progressive? I think that can safely go in the liberal group

settintotrieste:
Alright. I prefer to keep progressives where they are. They are not really liberal. They are like a hybrid between socialist/liberal so I prefer to keep them in the centre.

ARR8:
That's true, but the center implies a mix of left and right politics, while progressives are undoubtedly left. It seems more to me that liberal groups are the pro capitalist left, especially going off of vanilla, in which case progressive is a perfect fit. Either way, it seems logically that they would more willingly form a coalition with a liberal group than a moderate one.

settintotrieste:
Its 12.37AM on my side at the moment which is a bit late so I will need to get back to you on this issue tomorrow. Just to update you on my current thoughts, based on the Internet research I've just done, I'm thinking of creating a centre-left group for progressives as they don't really fit the description of center right nor center.

ARR8:
Sure, just keep in mind that, at least from NWO, progressives were meant to replace the vanilla "anarcho-liberal" ideology, as a far left minority ideology.

I think the ideological groups are best interpreted as parliamentary coalitions, and when thinking about it from that perspective, a green party would definitely sooner form a coalition with a center-left liberal party than a centrist moderate party.

That is also why I put populist in the conservative group - nowadays, you can see far-right parties entering coalitions with conservatives. However, nationalists generally have to work alone or with other minor parties.

settintotrieste:
I believe that it is better not to interpret them as parliamentary coalitions as may lead to biaseness. Not all countries follow the same setup when liberals cozy up with progressives or populists with right wingers. It is must more impartial and controversy free to link coalitions to the same section of the  political spectrum. With right parties allying with other right leaning parties, centre right with centre right etc. This should form a nice coalition spectrum that goes like this:

- Communitarian, community above state/Left wing (Socialist & communist & market communism)
-Less communitarian, more about individualism and statism/Centre left wing (Progressive)
-The right balance of communitarian, individualism and statism/Centre (Centre wing)
-Individualism and less statism/ Centre right (Liberal)
-Statism/ Right wing (traditionalists & conservatives)

Special ideologies:
-Nationalist - single issue party cannot be conventionally classified on the ideology chart
-Populism - This is tricky as populism is often misinterpreted these days to be far right or what not. The problem lies in the fact that all parties do practice populism as a party can only remain in power with the support of the people which will inevitably lead to the adoption of popular policies to keep the support. So what people say is a populist party can actually be a conservative/socialist party in disguise which can be seen in trump who many classified as populist but really in actual fact is behaving like a Neo conservative with traditionalist undertones that not many Americans subscribe to. So in order to ensure that the ideology itself receives accurate representation in the game, it is necessary to follow the fundamental definition of populism which is te following of the public will and the public will can be anything be it socialism, conservatism or liberalism. Right policies is only one area that populists can end up doing. So they should belong in the minor group.

At the end of the day, ideological groups must be distinctly defined according to their fundamentals and by extension their fundamental complements.

ARR8:
Overall, this makes sense, but I have a couple of problems with this interpretation.

First, vanilla interprets these groups as coalitions. Putting most parties in their own group defeats the purpose. Think of the activities of parties when trying to get a ruling coalition - a parliament which is 40% conservative, 30% liberal, and 30% progressive would always lead to a liberal-progressive coalition, even if liberal and progressive ideology are slightly different. At the end of the day, we should be grouping these according to these gameplay reasons, since this is the only capacity in which a player will ever see the groupings.

Also, a minor issue - I'm nearly certain liberals are not center-right, but center-left, with moderates almost always a compromise between mainline conservatives and liberals. Progressives, as the average Green Party, are to the left of liberals but not moderates.

Finally, I am in agreement on the populists. In thinking about it, there are several countries which have major parties refusing to form any coalition with populist groups, so it makes sense for them to need a higher majority to become the ruling party.  
However, my interpretation differs from yours in one respect - populist parties nowadays tend toward the interests of the rural or similar people as their base, whose interests are almost always right-wing and a reaction to huge changes which leave them behind, such as large-sclae immigration and outsourcing of manufacturing jobs. If the base changes to a more urban lower class, it will likely be better served by socialist populism, which is simply a socialist party. In the mod, looking at parties represented, it's almost always the right-wing sort. Perhaps they need a better name, but left populism I am sure is not part of the Populist ideology in the mod. If it is needed, it should be a new ideology.

Nationalism, I can also see being in their own group and only forming a coalition with other nationalist parties. Out of any, nationalist it seems to me has the most variety on left-right or economic issues. However, that can be represented in the party policy for each country.

In short, while I agree that in principle parties can differ greatly fundamentally, I think this behavior changes if they have a chance to serve their voters as part of a ruling coalition, and I am certain we should minimize single-party groups for the sake of gameplay.

PR #187, opened 2017-05-19
Tweaks, fixes, cleanup, strings

ARR8:
Nothing new here. Jerusalem kept as decision.

For East African Formation, unsure whether to automatically integrate Zanzibar, and not familiar enough with the history to make a decision. Have adapted the Tanzania events to also fire for East Africa for now.


settintotrieste:
Thanks!

Based on what I've read on the history of the proposed formation of British East Africa, Zanzibar was not one of the planned members of the federation so it should not automatically integrate Zanzibar.

ARR8:
Great - I might do some tweaking for the MTTH, but I think this implementation as a whole is fine.

ARR8:
In that case, should the federation ever integrate Zanzibar at all? Maybe it should have more difficult requirements than Tanganyika had.

The description for the event seems to imply a primary Anglo rulership, like Rhodesia, whereas the one in-game is primary Swahili, as though it were formed as a regular union. Should it mirror the Rhodesia events?

settintotrieste:
I would say yes but with the difficult requirements.

Yes, it should mirror the Rhodesia events where the primary culture becomes a settler minority when the federation is formed and changes to a black majority upon independence of the federation if it doesn't collapse by them due to ethnic tension.

PR #188, opened 2017-05-19
Size reduction 2

ARR8:
I think that's all identical files to vanilla removed. Also removed/replaced NWO templates which are now outdated/unneeded.

I'm not sure if they're enabled, but, if not, the vanilla music files could also be removed.


settintotrieste:
Thanks, but just to double confirm, it does not have any impact on the game itself?

The vanilla music files Must remain for the code to stop them from playing in a mod. This is because the vanilla music files when absent are taken from the base file instead.

ARR8:
No impact.

That's true for any file - the game executable is more likely to be corrupt than gfx text files, but we don't include that. I've not seen a single mod that includes unchanged gfx files. In fact, the gfx files are more likely to corrupt on download/extraction than the ones sitting in the Vic2 folder.

If I'm not mistaken, it isn't the songs.txt which gives instructions for the music? The music files themselves are identical to vanilla, which means taking them from the folder will have the same effect. Either way, this PR keeps them since I wasn't sure.

settintotrieste:
Alright but I prefer to keep the vanilla songs in the mod just in case.

PR #189, opened 2017-05-21
AI logic tweaks + Ethiopia rewrite

ARR8:
Something I've put in but am iffy about is having the issue of Trieste and Bulgarian unification autoresolve in favor of the player if no other players are nations involved. The way I see it, players are probably savescumming those events anyway. The Bulgarian unification one is temporary until the ^pending Yugoslavia rewrite, anyway.

I've read two separate complaints about the fixed nature of Ethiopian events, so I've rewritten them. Now, there's both a liberal and communist coup possible, with MTTHs based on the political climate of the country but otherwise equal. Only one of them will fire, and it is possible to avoid them by reforming into a constitutional monarchy or getting very lucky with the MTTHs until 1990. I've also rewritten some Somalia events so that the Ogaden becomes Somalian about 1/4th of the time. I've cleaned up Eritrean events to match the general changes we've been making to decolonization events, though there's one event I'd like to tweak in the future some more.

### DIFF
--- a/CWE/events/Essential Events 6.txt
+++ b/CWE/events/Essential Events 6.txt
@@ -198,7 +198,7 @@ country_event = {
 		year = 1956
 		NOT = { has_country_modifier = opecmember }
 		NOT = { year = 1990 }
-		OR = { capital_scope = { continent = mena } tag = VNZ }
+		OR = { capital_scope = { continent = mena } owns = 2266 } # Calabozo, VNZ
 		NOT = { has_country_flag = petro_rise }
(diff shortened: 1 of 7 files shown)


settintotrieste:
Please keep the federation of tanganyika and Zanzibar event unless a replacement has been created. It is my policy that all features removed must be replaced with one or more features such that there will not be a net reduction in features

ARR8:
There is a replacement - as far as I know, the rewritten events from before now do everything this one did

settintotrieste:
Ok

PR #191, opened 2017-05-21
UNSQ duplicate event fix and limits

ARR8:
Limits referring to only countries at peace getting the UNSG event and only UN members of vassals of such getting the global consciousness. Also tweaked the prestige values a little and minor cleanup.

Tested this - it does select all three kinds of secetaries-general and there is only one major event per selection.

### DIFF
--- a/CWE/events/Essential Events 4.txt
+++ b/CWE/events/Essential Events 4.txt
@@ -2361,54 +2361,51 @@ country_event = {
 	}
 	option = {
 		name = EVT_1412086_A # "Let's see who got the job"
-
-		random_country = {
-			limit = {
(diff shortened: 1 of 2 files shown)


settintotrieste:
I believe the previous setup is better as it allows the player to know what type of Secretary General was selected straightaway from the event headline. So I will have to reject this PR unfortunately.

ARR8:
Would it be better to add the word (Passive, Normal, Aggressive) into the title, or have it be its own line in the description?

ARR8:
I've just now added the type of UNSQ to its own line in the description

settintotrieste:
Great!

PR #192, opened 2017-05-21
Rhodesia rewrite and Somalia tweaks

ARR8:
Complete Rhodesia rewrite. For one thing, it is possible to play Southern Rhodesia without the game ending a few years in, now.

There's now a couple of possible outcomes for the Central African Federation. It can survive if it reforms in time, or survive without Malawi by seceding with more land. The UK can also influence circumstances greatly, and may choose to keep the white nation by force. If the UK were inclined to keep some part of Africa under direct control, this would be it.


settintotrieste:
Overall this is great! Though I must ask request if it's not already there to make sure that when CAF is dissolved, the new states still remain in the British sphere and that when Rhodesia changes into Zimbabwe it also remains in the British sphere to ensure historical accuracy.

ARR8:
I've made them friendly, so that Britain has the option of invading to restore status quo or defending them from a renegade South Rhodesia. Having them in the same sphere I feel would be a bad idea as it would prevent a player-controlled South Rhodesia from realizing a restored CAF. Either way, Britain should not take long to sphere them manually.

settintotrieste:
I agree. Perhaps you can make ZAM and MLW under the British sphere but RHO that has declared unilateral independence not under the British sphere. Then when RHO changes into ZIM, it goes back into the British sphere to reflect the reversion to British rule that allowed Zimbabwe to form? This should still allow Rhodesia to invade ZAM or MLW and allow the British to intervene.

ARR8:
That's a good solution, done.

settintotrieste:
Excellent!

PR #193, opened 2017-05-22
Resource icons from vanilla #108

ARR8:
For any good in the mod which corresponds to a good from vanilla, I've copied the vanilla resource icon in place of the one currently in place. This is partyly due to complaints that many of the icons are too similar and party from my own observations about the icons not fitting the rest of the game.

I've skipped all military goods except tanks, since the ones from vanilla would look pretty outdated. Some, goods which have no vanilla equivalent but are palette swaps in the mod I've made palette swaps from vanilla icons. Eventually, it would be nice to find suitable icons for all the other goods.

Screenshot:  
capture


settintotrieste:
Great!

PR #194, opened 2017-05-22
Slow populism rise from terror, fix convention multiplayer bug and static events, strings

ARR8:
Other countries' far right will only react to terrorism in your country if a secondary or greater power.

UN Terror Conventions set to fire for the UNHQ nation to fix the bug that only one player in multiplayer will get them, and their MTTH has been varied a bit.

Finally, string fixes and minor flow cleanup for one event.


settintotrieste:
This is alright but I need you to remove the major = yes for the UN terror conventions as an event with a picture is better than one without. I would have to stress that CWE does not support multiplayer compatibility if it leads to a loss of flavour in one form or another.

ARR8:
I think it would be better to forego the picture if it can allow every player to get the notification. We've gotten a lot of feedback from players using multiplayer, and I, myself, use multiplayer. It's a bit of a hassle to read the events out and I wouldn't argue much flavor is lost since the decisions have a picture, anyway.

Speaking of decisions, hold off on merging this for a sec - I've got to change those, too.

settintotrieste:
Ok I take your point.

ARR8:
Great, and the decision tweaks are now up for review, too. It's pretty much just a requirement of being a UN member, not being over the infamy limit, and also allowing any greater power who wouldn't otherwise meet the requirements (read: China) to join the interpol.

settintotrieste:
Great!

PR #195, opened 2017-05-23
XEF majority rule decision, Syria majority rule fix, strings

ARR8:
In lieu of a CAF-style event chain for East Africa, though that could be done in the future, I've added a decision to allow them to transition to black majority rule if they reform, get independence without nationalists, or take an example from other nations as the AI. This also grants the UK a CB to restore them as a vassal if eligible to use it, and unlocks the future union with Zanzibar.

Something I'm not sure about: Should taking this decision remove Zanzibar from the British sphere? Otherwise, I'm not sure it will ever meet requirements for the union, which involves Zanzibar not being part of a sphere. With Tanganyika, this is fine, since they get independence around the same time, but XEF can stay under white rule long enough for the British to sphere Zanzibar again.


settintotrieste:
Great!

I think an independent Zanzibar should be given a choice via event when XEF becomes independent to decide if they wish to remain with the British or to leave the British sphere to pursue greater ties with XEF with a possible eventual unification.

ARR8:
Good idea! I'll see about implementing that soon.

PR #198, opened 2017-05-25
Vassal decision cleanup, tweaks esp. potential

ARR8:
Vaccine random durations, switch economic focus, ai logic


settintotrieste:
There are some problems with this PR. As far as I know, random_list cannot be used in decision effects. At the moment I have some ideas on how to expand economic focuses to give more different flavours for both export focus and import flavours  so I request that it not be changed. The last thing is that I tend to like to keep the AI conditions in the allow box as I find it uneasy to have no allowed conditions being displayed. The allowed conditions also allow players to know how the AI chooses as well.

ARR8:
Keep in mind the decisions are intended to uncluttered the list for vassals with decisions they cannot take. So, in order:

You use the random_list for Space Race Attempts, and it's always seemed to work fine there.

Sure, I'll undo most of the changes for the economic focus. Currently, the change decision does not work, which is why I removed it. Are you planning on adding more than two foci? If not, I would argue that the change decision is unneeded as it only clutters the list, and the other decision can function as the change button when eligible.

The potential changes mean that the player won't see most of the decisions until they can take them. I would argue that seeing conditions in the allow button confuses the player unless they apply directly to them. I therefore remove the ai logic as a matter of usability.

Normally, I would agree BUT the player gets an event as a vassal setting UN Observer status and explaining how to become a full member. What I'll do is allow vassals who are almost secondary powers to see the decision.

settintotrieste:
If I'm not mistaken, random_list with country events only is alright but I'm not sure about country modifiers. If you can help me test it out.

Thanks, what I have in mind is to expand the economic policies to export, import and mixed, with multiple different options under each category to allow for better fine tuning of how the policies benefit the country.

I see your point and agree. Just make sure that the ai_will_do logic is correct as it's quite easy to make mistakes with the double negatives.

Alright, that's a good solution, please go ahead.

settintotrieste:
Thanks. this is good but may I request that all changes to the economic policy feature be reverted? I prefer the current layout as it is much easier for me to build upon it from there.

ARR8:
Sadly, the change economic focus in its current state is completely broken, and never allows actually switching focus. To expand upon it, you can simply put the conditions for whichever economic foci you add in an OR block on the change focus decision. Also, this one remains hidden until it can actually be taken and notifies the player when it's not.

settintotrieste:
Sorry, I have to insist on this as at the moment the presence of the decision is harmless and by changing this way it makes it more difficult for me to edit the economic policy feature. I can compromise by allowing the player but not the AI to toggle between the policies but I need you to revert the changes first or there will be a PR conflict.

ARR8:
Sorry, if I understand correctly, you want to decision to be present, but in its broken state?

settintotrieste:
No, I intend to allow the player to toggle between the decision but I need a reversion of the changes so there will not be a merge conflict.

ARR8:
I would advise, then, adding a commit to this branch because there are changes to the republic day decision also in that file. Although, to my understanding, I have done exactly what you plan to do with that decision.

settintotrieste:
Alright, I shall modify the PR.

PR #200, opened 2017-05-27
Tweaks to recent decisions

ARR8:
Vassals of OIA states can remain, open discourse needed for internet
rather than democracy, flags, strings, cleanup


settintotrieste:
Great!

PR #213, opened 2017-06-22
Update Energy Triangle.txt

ARR8:
Allow superstates to take decision and overall more dynamic  
Note I wasn't able to test this in game or in the validator, though it looks fine syntactically


settintotrieste:
Thanks, this is excellent work ARR8!

PR #218, opened 2017-06-23
Update version number to 1.06

ARR8:
In launcher and main menu


settintotrieste:
Thanks! There are just a few more final touches for me to make before I will release 1.06.

PR #219, opened 2017-06-25
Rewrite end of Korean occupation events

ARR8:
Rewrote the Korean occupation events to allow more dynamic outcomes and be aware of situations which may lead to either of the powers not releasing their respective parts of Korea, players of those countries releasing that land manually, and land in the peninsula potentially changing hands, such as after a war. Also includes some bugfixes:

- Fixed a bug which made it impossible to recognize the provisional government in the ROC
- Fixed some outcomes releasing South Korea with the wrong capital
- Fixed trusteeship not resolving as frequently as it should

Additionally, the new events from 5180a71 are now integrated into the rest of the event chain, and some other ahistorical outcomes were added:

- The Soviets may keep the land for themselves
- The Soviets may recognize the provisional government in addition to the USA
- The USA may recognize North Korea in extreme circumstances
- A democratic Russia may proceed with trusteeship without problems
- The ROC can unify Korea by installing the provisional government, even in the midst of the civil war.

Most of these outcomes will never happen under the AI, but this can be changed by tweaking some base AI chances.

No events were added. I only added more logic to existing events.


settintotrieste:
This is excellent! I approve of the changes though I've done a small tweak in the fall of the USSR event to not release KOR as the event will almost always fire when the USSR collapses so causing Korea to be reunited far too often.

ARR8:
Thanks! I considered doing it that way, but decided against it. I thought it was a little unclear, so I left that comment. Basically, KOR will not have cores unless NKO does not have any, either. That way, only one of them will be released. There is an option along the chain to replace NKO cores with KOR cores, so keeping those lines in may be necessary or Korea won't be released at all.

However, on review, that portion of the code is faulty, as you say. I'll add a commit that should fix both problems.

settintotrieste:
Great!

PR #222, opened 2017-06-26
Supercomputer and Visegrad cleanup

ARR8:
AI, more nations to Visegrad, Visegrad relations effects

A question: are the prestige effects for supercomputers supposed to be so high? For an amount of money which would be relatively simple to accumulate for the player, it seems like a lot to get.


settintotrieste:
Great work!

Yes, as usually at that late stage of the game, the prestige score of the big nations are rather big so an addition of 10-50 prestige is not sufficient enough to make an impact for the scale of what is achieved.

PR #223, opened 2017-06-27
Add Weihai as SAR, Guangdong cores

ARR8:


settintotrieste:
This is interesting, may I know more about this update?

ARR8:
Sure - I've added cores to a new potential SAR of Weihai. In real terms, it probably won't appear in-game unless released. However, it is a former British colony, and a possible future addition could be an Imperial Federation who wishes to take back all provinces of the empire may get the chance to re-vassalize Weihaiwei.

I'm also planning on adding Qingdao, a former German colony in the same state.

settintotrieste:
I see, this is a good addition!

PR #225, opened 2017-06-27
ManCan civil war fixes

ARR8:
If the Manchurian Candidate chain were used on a nation with all of their cores also the cores of another nation, such as a not-yet-formed superstate, the first nation would be completely dissolved. This is fixed.  
It may be advisable to replace the code for releasing countries with the code for dismantlement at some point, as this already avoids the issue by leaving the capital state intact. However, this works fine for now.

Another problem with the chain would arise if the UN headquarters were to be moved, but all of the eligible provinces in Europe are part of a superstate dictatorship. Two more possible locations for the headquarters have been added, one in Asia and one in Oceania.

### DIFF
--- a/CWE/common/event_modifiers.txt
+++ b/CWE/common/event_modifiers.txt
@@ -327,8 +327,9 @@ joint_aid = {
 	tax_efficiency = -0.05
 }
 
-### Religion Flag ###
+### Flag ###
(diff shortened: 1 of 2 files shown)


ARR8:
As part of some future larger work on this, I'd also like to add Nairobi, as it turns out there is a large UN building there already

PR #227, opened 2017-06-28
Set some decisions on game start

ARR8:
All UN members, Japan's right to war, and Soviet centralization

This also sets some members which can not join under the current mechanics, such as Cuba, who was a member but is represented as a vassal, and Ukraine and Belarus, who will automatically be members on independence as their respective SSRs joined the UN.

Something that may be wise to check - I made the Republic of China a UNSC member on game start, as it was by 1946. Hopefully, there exists something to remove this when it becomes Taiwan. If not, it can be easily added.

If this works well, I'll probably see about setting more things like this. I particularly like automatically renouncing the right to war for Japan, as otherwise there is no reason for the player to do so.

### DIFF
--- a/CWE/history/countries/ARG - Argentina.txt
+++ b/CWE/history/countries/ARG - Argentina.txt
@@ -46,6 +46,8 @@ social_service_reform = no_social_service
 # Starting Consciousness
 consciousness = 5
 nonstate_consciousness = 0
+decision = united_nations
+decision = jointheunga
 schools = army_tech_school
 oob = "/ARG_oob.txt"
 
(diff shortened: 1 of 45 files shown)


settintotrieste:
Thanks!

The removal of Taiwan's UNSC seat has already been implemented.

ARR8:
Something I am considering is presetting centralization to correspond to certaion [unitary states and federations](https://en.wikipedia.org/wiki/Unitary_state#/media/File:Map_of_unitary_and_federal_states.svg), and leaving them at the player's discretion to change. However, I'm not sure how to handle historical changes in this status without adding very specific AI logic. Just an idea for the moment.

settintotrieste:
This is too difficult to implement so we shall keep the current set up.

PR #231, opened 2017-06-29
Fix #229, fix #214, nation of equals for ETH

### LINKED ISSUE #214: Transfer of the Ogaden complex conditions
f4f62a3

I'm considering this a bug since the conditions were designed to keep the Somalian option infrequent 😄.

For the record, the British were in all likelihood going to grant Somalia the Ogaden if it weren't for pressure from America - so it should definitely happen sometimes.

### LINKED ISSUE #229: Ethiopian state review
Guest:

Couple of things that may need fixing: A number of Ethiopia's states are shaped really weridly, e.g. Begemder is in two parts on opposite sides of the country. Not sure if that's intentional or not.

ARR8:
#214: This was caused by a bug in one of the Somalia events, causing the formation of the Trust Territory to include the Ogaden. If the Ogaden event fired first, this would not happen, even if the territories were kept, due to Somaliland cores being added. In any case, an exception was added and the MTTH tweaked. The complex conditions from before were restored as they were unrelated to the problem.

#229 before and after:
v2_18
v2_19

Some other map-related bugs were fixed in this area, such as names and positions.

Finally, the decision name for revoking the nation of equals act did not make sense for Ethiopia in 1946, who starts with an accepted culture, so it was changed.


settintotrieste:
Great, thanks!

PR #61, opened 2017-03-23
Italian parties changes

KeyboardMonkey95:
I added appropriate parties for both starts and made some other general adjustments, including reviewing their policies and time availability.

PR #243, opened 2017-07-07
Italian parties translation + small bug fix

KeyboardMonkey95:
_#241_ - ARR8


ARR8:
Thanks! By the way, if you're as knowledgeable about Italian local politics as you are about Italian national politics, please feel free to do a similar rework as #61 for any of the Italian separatist states, which are pretty lacking.

KeyboardMonkey95:
Ok, i'll see what i can do, but the problem is that separatist movements aren't really a thing in Italy. Because, even though there are some very minor groups who are serious about it, most politician in those contexts want more autonomy rather than actual independence. 

Even the "North League"(Lega Nord), which in the '90s and early 2000s wanted Northern Italy to secede to form "Padania", nowadays doesn't seem to care about it at all.

PR #246, opened 2017-07-15
Rewrite of Austrian occupation events

ARR8:
The end of Austria's occupation now has many more options. The events start with the USSR, as before, and the allies can respond to any decision the USSR makes. Notably, it is now possible to divide Austria like Germany, and America can bolster West Austria by merging it with Bavaria. It is also possible for the player to accept a united communist Austria, but the AI never will. The behavior from before was also kept as an option, and it is possible for any player at any step to decide to keep their Austrian land permanently, for infamy. They will periodically be pressured to surrender it, however.

The proposed option to allow a communist Danubian federation wasn't added yet. Another planned addition is a claim greater decision for the SGF, although minor conflict with Germany over Kaiserslautern was added for now.

There are a few TODOs in the file, and testing is needed to see how it plays out in-game. I checked that the events work and that the AI uses them appropriately, but I have no clue as to balance or how anything might play out through the engine and outside of events. I've put what I felt were reasonable AI base chances, but those are open to tweaking. I've set the USSR to strive for neutral Austria 65% of the time, and communist Austria the other 35%. Infamy is also something that should be tested - it was mostly untouched here. Although that was supposed to be the point of the rework, the Soviets will likely lose enough infamy over Austria in the 50s that any manual release probably won't be worth it.

Overall, this should lead to a very dynamic situation in Austria. Divided Austria will never lead to war, but a softening of position of the part of the Soviets, leading to a united, neutral Austria (sound familiar?). The SGF will be created 1/8 of the time the Soviets attempt communist Austria, or about 5% of games.


settintotrieste:
Excellent work!

Issue #250, opened 2017-07-17
Suggestions and improvements for The Netherlands and Benelux


DutchDylan:
I was doing a playthrough of the Netherlands in which I went on to form the Benelux, but during my playthrough I came accross various stuff which could potentially be changed for a better experience.

First of all, I noticed the implemented Benelux seems similar to the one discussed over at the [NWO mod page](https://github.com/simsulla/nwomod/issues/473). Which I think is very cool since I had a role in shaping what the Benelux would look like ingame. But during my playthrough of the Netherlands into Benelux I noticed a few things which could be changed to improve this specific part of the CWE mod.

So let's begin, when I played as the Netherlands not every party had a unique name, as well as some parties not having similar policies compared to the parties itself in the Netherlands. I gave a unique party names to every party, but since some ideologies are not as well-present in our political landscape as the others I gave them the name of a Political party which would be closest to them. 
I've done the same for the parties for Benelux union, basing them on existing parties in the Netherlands Belgium, and also Luxembourg.
As a minor thing, the ruling party in both starts should be the big tent one, since it is the one most similar to the party then leading the coalition.

The political reforms at the start also weren't totally accurate, The netherlands uses the D'Hondt for voting, which is called the Jefferson Method ingame, but First Past the Post is selected instead. During election showing up at the polling station was required but voting was not, in 1970 this was scrapped and showing up became voluntarily.
So it would be better if at the 1945 start the Netherlands has the Jefferson Method and the Support Disenfranchised selected, and has the Jefferson Method and Universal selected at the 1992 start.

Then for the map/provinces, province 380 is called 'Flevoland' but should be named 'Breda' instead. I assume this error is in place because the previous 380 was merged with 379 to create a single province and create a new empty prvince for the Flevoland province which would be the drained Zuiderzee. Which was previously implemented in the NWO mod. 
Then for the region which 380 and 379 are in, I propose for it to be named 'Brabant' since that region primarily covers the territory of what would be the province Brabant in real life.
The province of Luxembourg, 397, should be part of the Walloon region. For the country Luxembourg itself it doesn't matter which region it is part of since it's an OPM, but if forming the Benelux you would have a single province to manage for factories and stuff instead of it being part of a region with multiple provinces (that you control) which would be more practical, from a historical perspective it could also be said that Luxembourg would have been administrated as part of the Walloon region during the rule of the Kingdom of the United Netherlands.

[...]

PR #254, opened 2017-07-18
Benelux stuff

DutchDylan:
These are the first implementations of the mentioned suggestions in #250.

For now there have been added two events upon formation which let the player pick a capital and monarch, another event has also been added which triggers upon the nationalist party getting elected resulting in the end of the union.
Minor adjustments have also been made in the already existing decisions and events. To create compatability with the new monarch event, readding German as an accepted culture since a primary culture can't ne changed, as well as a cleanup to give more readablity when forming the Benelux. But also to have a stricter requirement to forming the union, so belgium has to be unified and can't be split between Flanders and Wallonia.


DutchDylan:
Thanks for the compliment.

My line of reasoning was that since the Benelux was an initiave between the three nations, Netherlands Belgium and Luxembourg, they should be the ones to form it. A divided Belgium could already imply that co-operation between the Flemish and Wallonians (or Dutch and French to be blunt) is rather impossible, so why would they try again on even a larger scale where their interests would be harder to be met since their part of the population would be even smaller resulting in less politcal power. This being said from a Flemish perspective, if grouped together with the Dutch they would be a steady majority, but the Flemish and Dutch interests would not always align, even if the common language would unite them so to say.
And about the Flanders Wallonia split after the collapse of the union, I, uhm, I kinda forgot to implement that.. Will do it with this revision. The only problem is that if a player started as Belgium, that after the split they can't switch back to Belgium. Perhaps I should leave Belgium whole in that specific scenario, so it gives a bit of uniqueness to the event like with the specific Dutch version of the event.

- I Hadn't encountered these modifiers before, I will look into how to use them and replace the old modifiers with these ones.

-Will do.

-Ah yeah makes sense, I forgot to realize a democracy gets that event too so it needs a switching.

-Very interesting, I haven't worked with that before either so I will look into it how to use it. The current weights implemented are just speculation of what would be most realistic.

-And for the last two, I will work on my bookkeeping. I personally used the tags instead of THIS since its clearer to me, but I will try to adhere to the modding style of CWE.

And yeah I tried that too, but as far as I know you can only check through the trigger part before the effect part. So I dind't think there could be a universal event which triggers under a circumstance, and has effects based on varying differentiating circumstances.

DutchDylan:
@ARR8 Are your sure about the AI chance being less than 1? I've looked into it and other files within the CWE mod seem to use a scale from 0 to 100. For example, the 80, 10, 10 split in the events file for Belgium. Might it have been that you meant [the factor within the modifier](http://puu.sh/wNOEX.png) in the ai_chance itself?

ARR8:
Yes, the modifier actually can be >1; see #117. To save processor cycles, the AI will automatically take the first choice that's 1 or greater. Events that have it the other way are left over from NWO and need to be changed at some point.

DutchDylan:
Interesting, didn't know it could have such an effect on performance. 

As stated in the commit I checked up the three Benelux countries and cleaned up the ai_chances there as well.
The various mentioned gripes have been fixed except for two. The events for choosing a monarch and capital don't acount for starting/forming nation, and at the breakup of the union Belgium still gets released whole. For now I would like to leave it at that and implement these two features later on, since I'm not sure how yet how to implement these two correctly. Specifically I'm not certain how the modifier within the ai_chance influences the ai with choosing an option in an event, and I'm looking into how to combine the several breakup events into one for effeciency.

settintotrieste:
All is_triggered_only = yes events must not have the trigger = { ...}

PR #259, opened 2017-07-30
Belgium focused enhancements/corrections, minor Benelux enhancement

DutchDylan:
A minor event was added for the Benelux which fires upon formation. 
The event gives the player the choice of how to represent the country on the sportsstage with minor effects on people's ideology (and prestige loss) for the options of the event. Primarily aimed at people who roleplay in their playthrough, the minor effects are there so a none-roleplayer has also something 'valuable' to choose form, hence the minorness.
The names were also changed so the union is removed from it since Benelux can be used as a standlone (eg compared to the EU, were 'European' would not indicate a country by itself and the union party is necessary).

For Belgium the parties have been reworked, currently the policies of the parties haven't been changed but only the names. Each party was given the name of a currently existing party for the appropiate region, so Flanders and Wallonia each have their own party like irl. The only exceptions being the communist party, its name originally has Belgium in it but it's weird for there to be two 'PVDA Belgium' parties. This resulted in replacing Belgium with the appropiate region (Flanders, Wallonia). The ideologies which lacked a party were given a name combination of the ideology and the associated culture/nationality.

The two regions parties have also been reworked, as in their names now copying over the names of the Belgium parties for their appropiate region (using _0_BEL and _1_BEL instead of _0_FLA and _0_WLL). In the future if there were to exist an independent city of Brussels (I volunteer to work on the necessary files :P), it could also reuse the already existing party names.
The starting reforms were fixed for Belgium and the Netherlands to represent the compulsory attendance for voting existing in 1945, and Belgium still having them (officially by law, in practice not anymore) in 1992.


settintotrieste:
This is great however there is a minor thing to take note of. The code to increase infamy is badboy = X and not infamy = X

It would be better to keep the name The Benelux Union and The Benelux Kingdom as it would be strange in real life and to the average player to have a country called just Benelux.

DutchDylan:
Ah damn that I didn't notice the badboy thing, I will change it and revert back the naming.

settintotrieste:
Thanks!

DutchDylan:
btw, what are your thoughts on a 'free/international city of brussels' similarly to the ones of Berlin and Jerusalem already in-game?

settintotrieste:
It sounds interesting and I wouldn't mind having this but it will need to tie in with an event or decision chain.

DutchDylan:
Yeah makes sense. I haven't though of a specific situation under which it would come into existence, only that it would be due to the breakup of Belgium, so I will keep it as a backthought for now.

PR #262, opened 2017-07-31
Luxembourg political party naming, starting ruling party fix

DutchDylan:
Naming of Luxembourgish parties, and other minor fixes.


DutchDylan:
Also sorry for having a pull request so soon after the last one, I had expected this one to take more time than it did. Otherwise I would have done them together as a single request.

settintotrieste:
No problem, this is good.

Issue #204, opened 2017-06-07
Decisions to merge divided nations


ARR8:
Like the decision for Korea, it may be desirable to create similar decisions for Vietnam or, for a Turkish or Greek GP, Cyprus (or potentially as an atheist dictatorship, e.g. communist, as religious differences would be ignored). Also possible for Romania/Moldova/Pridnestrovye if the GP is not Russian culture, also Finland and Karelia, etc.

ARR8:
#250:
> -If a great communist power like the Soviet Union or the Chinese People's republic has sphered all the three low countries and all three of them are communist dictatorships already, then there should be the possibility for that great power to unite the three through a decision. This should only be the case for extremist dictatorships and not for democracies and the like.

settintotrieste:
Some decisions were made in the end

PR #266, opened 2017-08-08
Dutch Political Parties - Historical Enhancement

DutchDylan:
Political parties were added and modified for the Netherlands.
Upperhouse seats corrected.
Religous populace representation fixed.


DutchDylan:
Stuff as previously mentioned in #250 

Sadly I haven't found a way to modify starting ideology yet, my backup-plan is having an event fire at the start to change/fix it but I will implement this later.

As a next enhancement I would like to mod in the [Antilles](https://en.wikipedia.org/wiki/Netherlands_Antilles) as country so it can be used for an event chain which will fire on appropiate circumstances. Since a new country will be added by modding this aspect I would like approval first before I start on it.

DutchDylan:
Implemented decision to form Benelux under a Communist government as mentioned in #204

DutchDylan:
As mentioned in the commit, replaced two flags because they could use some flavour and were kind of unfitting.

settintotrieste:
Thank you for your contribution! This is excellent! 

Netherlands Antilles has already been added into the mod. But currently it appears by event which only fires in 1946 start date to establish a Netherlands dominion called Antilles. Antilles' tag is CUR which it shares with Curaçao. The name Antilles will appear when CUR is a colonial_government.

DutchDylan:
Ah I see, it never appeared in any of my games/runs so that's why I made the suggestion. I will check out how Antilles is currently implemented and perhaps make some tweaks.

settintotrieste:
This is the event 1412166

PR #273, opened 2017-08-16
Added parties for the Italian separatists States

KeyboardMonkey95:
I also tweaked the Italian parties to avoid the same ideology overlap.


settintotrieste:
There is a bug with this update causing a crash adapting history when selecting a country so I have reverted this PR until it has been fixed.

PR #325, opened 2018-01-03
Accepted cultures of Caucasus republics

RyongJin:
Updated the accepted cultures of these republics to be more historically accurate. The official languages are shown respectively in these wiki links.
  
https://en.wikipedia.org/wiki/Karachay-Cherkessia
https://en.wikipedia.org/wiki/Kabardino-Balkaria
https://en.wikipedia.org/wiki/Chechnya
https://en.wikipedia.org/wiki/Ingushetia

If it is okay, I can work on a "de-russification" decision for all caucasus republics in order to remove Russian as an accepted culture if any of the republics become independent/get a nationalist or populist government.

### DIFF
--- a/CWE/history/countries/CHC - Chechnya.txt
+++ b/CWE/history/countries/CHC - Chechnya.txt
@@ -1,5 +1,6 @@
 capital = 1055
 primary_culture = chechen
+culture = russian
 religion = secularism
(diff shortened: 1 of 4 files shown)


settintotrieste:
Thank you for your contribution!

Sure, please go ahead and make those decision chains.

PR #326, opened 2018-01-06
Derussification Caucasus

RyongJin:
Added a political decision for all Caucasus republics to be able to remove Russian as an accepted culture in a derussification process.


settintotrieste:
Thank you

PR #339, opened 2018-02-18
Kayah state

RyongJin:
Added the Kayah state (copy paste from Kayin) and made provid 1338 its capital.

I also changed Taunggyi to Loikaw since Taunggyi is actually located in the Shan state (provid 1340), and Loikaw is the capital of Kayah state.

https://en.wikipedia.org/wiki/Kayah_State
https://en.wikipedia.org/wiki/Taunggyi
https://en.wikipedia.org/wiki/Loikaw


settintotrieste:
Thanks but did you forget to add in the country history file for the new state?

RyongJin:
Ahh sorry about that forgot to upload the file.

PR #359, opened 2018-06-05
Add files via upload

Anonymous1111111:
Better mercosur flags.


settintotrieste:
This is great! However the game engine only allows 1 communist flag per country so I would need you to remove the communist2 flag. Thank you for your contribution!

PR #409, opened 2019-03-02
Update some cold war events

nerd73:
One change is to stop the USSR almost instantly sphereing Yugoslavia post Tito-Stalin split.

The next set of changes have to do with what ends the cold war. America becoming an authoritarian (non-communist) dictatorship would not end the cold war, if anything it would intensify it as they would no longer have "internal traitors" to deal with. However, what WOULD likely end the cold war in a soviet victory would be the election/revolution of left-wing parties to power in the US, as they would likely seek cooperation with the Soviets.


settintotrieste:
While I agree with you that a right wing dictatorship won’t end the Cold War, I don’t agree that a full democracy with a socialist/communist ruling party would end the Cold War because the party can easily lose the election the next round to a right wing government who will then continue the Cold War. The UK and France were run by socialist governments at one point or another in the Cold War and they weren’t that friendly with the USSR to end the Cold War. 

I read somewhere that the Cold War ideological battle was between liberal democracy & capitalism vs authoritarianism & communism so a victory in the USSR’s favour must be a long term weak or an authoritarian and left wing USA.

nerd73:
That's also a good point, perhaps the US has to either slip out of top 2/3 or both not be a democracy and have a socialist/communist/market communist ruling party?

PR #411, opened 2019-03-04
Add  hms_government absolute_monarchy Iraq

bootyeater420:
The Iraqi monarchy was dominated by Sunni Arabs as such should be represented in the change primary culture decision


settintotrieste:
You'll need to add an OR = { government = hms_government .... } for it to work or it will check for a country with hms and absolute monarchy government type which won't exist.

bootyeater420:
Sorry a little rusty when it comes to Modding

PR #412, opened 2019-03-10
Update eurasia.txt

JohnLockers:
Allow mena region nations to join Eurasia, Eurasia join decisions now only require nation to be sphered by Eurasia or Russia


settintotrieste:
I want it to be such that only states bordering Eurasia or Russia can join. This is because my impression of a Eurasian superstate is one where it is a continuous bloc.

Issue #330, opened 2018-01-14
Trade Block Ambition for nationalist/dictatorships


JohnLockers:
I noticed that Eurasia and other authoritarian trade blocks are difficult to enter for the reason that the "An Open Market - Trade Block Ambition" decision is required. This decision requires a free trade policy and an election, or market communists to be in power. Perhaps authoritarian unions should be able to skip this step, or more options could be added to compensate for this.

PR #413, opened 2019-03-10
Update eurasia.txt

JohnLockers:
Remove requirement of flag trade_bloc_ambition - allows dictatorships to join easier issue #330


settintotrieste:
The flag is needed to prevent Eurasia from annexing the country if its in another trade bloc.

PR #414, opened 2019-03-11
Preliminary occupation zones

nerd73:
This patch adds the occupation zones (Germany, Austria, Manchuria, Korea) as puppets with a special government type. It also updates the establishment of those countries to reflect these new additions.

Occupation Zone governments have a -0.2 modifier for prestige and a -100% modifier for diplomatic relations. They are also forbidden from joining the UN in any form (both observer and full).

Another change is that the Soviet puppet governments (besides the zones) can now join the UN fully.


settintotrieste:
May I know the purpose of changing the current system? 
I’m not in favour of adding 400 plus flags, a new government type just for a feature that accounts for only 5 years in game. I would be better, in my opinion, to just create separate tags for the occupied territories with an event to mark which is which and another event to change_tag in the event that any accidentally become independent. Then edit the events to reference these nations.

settintotrieste:
Dynamic occupation zones can be modelled via the existing mandate government type.

settintotrieste:
Redone the occupation zones mechanic

PR #415, opened 2019-03-15
Politics

Glariamn:
Added ideologies of Social Democracy, Social Liberalism, Fascism and National Conservatism
Strengthened Socialist Movement
It works, but is mostly just there for being fancy
It is your decision if you choose to include it
If yes work will be continued and everything added needed for it to be fully incorporated into the mod


settintotrieste:
Thank you for your contribution however I’m not keen to add these new ideologies in because it will involve a massive amount of work to add them in properly. Adding to the current list of ideologies will involve adding new parties to every country and redoing all events and decisions to reference these to avoid bugs like repeating decisions.

Outta-my-way-burrocrat:
@Glariamn would you like to start working on this old CWE politics update again? The code history can be recovered from this PR using `git fetch upstream pull/415/head:glariamn_politics_recreated` on CLI

ccfox:
All of those ideologies aside from social democracy are adequately represented now. Victoria II can't really handle a whole lot of ideologies, both for optimization reasons and practical considerations with how ideology works in game.

Glariamn:
The political aspect of the game has more in common with Samuel Huntingtons rubbish or stuff like Mackinders Heartland theory than to more orthodox views on politics imho.

Not to say that it's the fault of CWE (which tried to remedy much of it) rather than the original mod it built upon (that had rather problematic conspiracy stuff associated with the far right) and i wouldn't touch upon it again if possible. 

Idk how much has changed, but from the looks of it fantasy aspects like irredentism are still in the mod instead of many irl claims and border disputes that could have been.

The political change i made 4 years ago was rejected even if it was an attempt to replace some of the political ideologies (looking at you big tent, market communism and populism) with something more universally acceptable.

ccfox:
Slight correction, the mod CWE is based on - nwo - is not the 'original mod'. If I have the order right, it was built off of a modern day mod, which was built off of a WWII mod, which was built off of an even older mod that itself incorporated other mods. Part of why its a mess is that, there was never an effort to deconstruct it, and now its kind of too late to do it for CWE2 - hopefully for Victoria 3 it will be better. However, even still, since I became more active in helping out with the mod I've been trying to target and remove all of that stuff - though there's a lot of it, I don't have much time to scout it out, and people only sporadically point out specific things.


Speaking to explicitly the ideologies - Victoria 2 can't really handle that much many more. The game engine is pretty dicked. More ideologies not only means more lag, but also that the ideologies will not function as fully intended. Pops kind of drift from ideology to ideology, with hardcoded elements making it even more difficult to work with. Hopefully when paradox ceases to add feature creep stuff that doesn't matter - like a focus on individual characters - to Victoria 3 and decides to flesh out the politics we can have a more workable ideology system there. I've already made proposals on the server that people seem to enjoy, though I still try to utilize the method of attempting to keep things simple.
________________________________

PR #418, opened 2019-04-01
United States election event revamp

nerd73:
US flavor events were railroaded regardless of election event choices. (such as Nixon resigning from the presidency even if humphrey/mcgovern are picked). This fixes that and also improves the events in minor other ways, and gives the option to disable them. If they're disabled, the historical path is taken with regards to political flavor events.


settintotrieste:
Thanks, this is good however the ai chance factors must be in 0.X for X% as anything more than 1 will be regarded by the game as 100% and it will select the first option with 100% chance all the time. I noticed that some of the election events also don’t have a check for the disable US election rigging flag. Lastly I feel that the disable US election rigging decision should cap its appearance to the last election event (eg. year = 1990) so it doesn’t appear anymore once all events have fired.

nerd73:
The election events that don't have a check are just later parts of event chains (like the watergate event chain)

PR #466, opened 2019-12-12
No more RICE economy tweak.

Etaglibash:
Erased rice form the face ot the earth. Changed every reference to it to grain. Even the good icon.
Tweaked factories so that they need electric gear, and other changes in the maintenace. Oil is now much more important to the economy.
Changed basic factories like machine parts, electricity and electric gear so that they dont need the same goods they produce (electricity plant makes its own power, so they dont need it for its own maintenance).
Tweaked beverage, canned food and alcoholic beverages factories. 
Tweaked pop needs.

I addresed your concerns settintotreiste. Changed building costs and factory production.


Etaglibash:
It NEEDS map cache erasing and new game to work.!! Wont work on save games!

settintotrieste:
Thanks, I like your work however I need to request a few changes before I can add it in.

Can you revert the factory building cost to the original amounts? Based on my previous testing, this amount tends to balloon to extremely large amounts for communist states which sucks up all the resources in the world eventually.

I would advise against increasing the factory good demand or outputs too much because by the end game, factory levels tends to rise to around +300 which pumps too many goods into the game causing extreme lag. Also, the earlier amounts were balanced to keep all base profits profits around 50% to avoid encouraging the ai to spam the most profitable factories.

Etaglibash:
Hi settin. I addressed your concerns in this new commit. I wasnt aware that factory buildong costs were that important. I lowered them all but kept the electrical gear costs. They are modern factories after all. I lowered the outputs I increased. 
My concern about that is that some factories seem never able to be profitable, even if their goods are demanded. Specially clothes, footwear and canned food. Need more testing on this.

Etaglibash:
Changing back the fabric and clothes and footwear factories. If they produce less, they are not profitable even if the product is at 20$ and they trash the entire economy.

settintotrieste:
The changes look alright. I'll add it in

settintotrieste:
I noticed that everyday clothes is listed under industrial goods, could you help me to move it back to consumer goods?

PR #468, opened 2019-12-13
Artisan Overhaul + localisation fixes

Etaglibash:
Also fixed the clothes good being in the industrial category. Switched it with electric gear, wich was in the consumer goods category.


settintotrieste:
Thanks however I cannot incorporate your localisation file directly due to problems with github encoding csv files. 

You can however list down the localisation changes you made and I will add it in for you.

PR #477, opened 2019-12-24
Various econ fixes, factory bonnuses and some event fixes

Etaglibash:


settintotrieste:
I approve of all changes. I noticed that you removed the has_country_flag = deny_superstate requirement in the mesoamerica decision. That flag is required to allow sphere leaders to decide if their spherelings can join local trade blocs.

PR #481, opened 2019-12-29
Update regions to better represent the post war border

HKKNNT:
-update China's states(regions) to match mordern China more
-update some South Asia states to better represent the post war border
-fix map positioning of 1563(Jinan)


settintotrieste:
Does the number of old regions more or less match the new number of regions? I have a rule to avoid reducing regions to avoid impacting the economy of the country due to the 8 factory per state issue.

HKKNNT:
Updated
Numbers of regions didn't change 
Didn't change any with East/West North/South too
Just removed those with strange names

PR #486, opened 2020-01-04
India, China interface update

HKKNNT:
-map update for western part China
-merge India's North Bengal region to West Bengal region, create Ladakh region
-remove Tibet region for India, create Arunachal Pradesh region
-reorganize Northern Kashmir, Eastern Kashmir, Western Kashmir to Kashmir region and Ladakh region, same region numbers for all involved countries
-minor change for Greater Tibet decision
-name update for Tajikistan
-interface update for POP screen


settintotrieste:
Thanks however Aksai Chin needs to be in its own state to prevent China and India from grabbing unwanted territory if they try to seize it though CB.

HKKNNT:
updated

PR #489, opened 2020-01-11
1 New province

HKKNNT:
-add Aouzou Strip in Chad for Toyota War between Chad and Libya
-update region setting for Chad to post ww2
-update naming of Nigeria's region's
-fix xec_enlarge_union


settintotrieste:
Thank, I like it. Could you help me make a province for Anguilla as well?

PR #504, opened 2020-01-30
Update graphicalculturetype.txt

Zaplord89:
Wanted to know about this


settintotrieste:
From what I know, there is no JewishGC

settintotrieste:
GCs affect unit Sprite type if there is one for that graphical culture in the files

PR #514, opened 2020-02-07
Small fixes and tweaks to some European events

Radical-Rust:


settintotrieste:
Thanks, however ai_chance needs to be within 0 to 1 to function correctly as the game will take any value above 1.0 as 100%

Radical-Rust:
Didn't know that. I'll fix everything right away.

Radical-Rust:
Fixed every instance of ai_chance were the value was higher than 1.

PR #517, opened 2020-02-09
Bugfixes

Radical-Rust:


settintotrieste:
Thanks. Some minor changes are required though, the extra bracket for the WMD should be added below is_vassal = no. Hong Kong must also be a non colonial state in both 1946 and 1992.

Radical-Rust:
I'm trying to edit 1946 Yugoslavia's pops file to adjust provinces pops, but no matter what I do, whenever I boot up the game Yugoslavia's total population drops to 0.
I feel like I'm missing something. Can you tell me what I'm doing wrong?

[Yugoslavia.txt](https://github.com/settintotrieste/Victoria-2-Cold-War-Enhancement-Mod-CWE/files/4177585/Yugoslavia.txt)

settintotrieste:
The province for Koper is missing a closing bracket

Radical-Rust:
I checked extensively with Notepad++, there's no extra bracket in the file. And even if it were, it wouldn't make the whole population disappear (I tried doing it purposefully with another country, nothing major changed).
The only way I could replicate the bug was by removing another country's pop file from the folder; I assume that, for some weird reason, the game isn't reading Yugoslavia's file.

Radical-Rust:
I ended up copying everything into a new txt file, switched it with the old one, and now everything runs smoothly. Weird.

settintotrieste:
Strange, there must have been some corruption

PR #527, opened 2020-02-23
Edits to tunnels and bridges decisions and events

Radical-Rust:
- Adjusted "Gotthard Tunnel" decision and event to match the Seikan and Channel Tunnel consrtuction chain;

- TIdied up "tallest_buildings.txt", and moved tunnels to separate (already existing) empty file.


Radical-Rust:
Is there any particular reason only tunnels take time to build, while other structures such as towers and bridges are immediate?

settintotrieste:
It was an experiment I used to see how to implement nuclear development. I didn't extend it to the others because it would unnecessarily complicate matters.

Radical-Rust:
I see. Then shall I make tunnels decisions like the other, or do I leave them as they are?

settintotrieste:
I prefer to keep the existing system since tunnels take a much longer time and costs a lot more to build than the other projects. Also I prefer to avoid making too many one time only modifiers.

PR #531, opened 2020-03-01
New West Germany events

Radical-Rust:
- Added new events for West Germany that deal with the rise and fall of political terrorism from the late 60s to the 90s;

- The Olympic medals event will now happen at the end of the Games rather than at the beginning;

- Added missing localisation to the years_of_lead modifier.


settintotrieste:
Looks good. There’s only change required: it should be country_event = { id = 123419 days = 16 }

Radical-Rust:
Should be fixed now.

PR #539, opened 2020-03-05
New buildings decisions, and other changes

Radical-Rust:
- Added 11 buildings decisions, and edited the requirements of the other ones. The new buildings are a few bridges, plus a new "category", flood barriers, which give the province a minor rgo and life rating boost (I only added the largest or most important flood control projects, like Netherland's Delta Works, London's Thames Barrier, and Venice's MOSE );

- Added a new event for France, the 1961 Vitry train bombing. It will only fire during the Algerian Independence war, or if Algeria is still a French satellite;

- Changed Siena's trade good to gas. The area is home to a large number of geothermal power plants, notably the Larderello station, the oldest and only such plant in the world until 1958, wich currently accounts for 10% of the entire world supply of geothermal electricity (Italy is seriously lacking in natural resources, so this is a small way to fix that without unbalancing things too much).


Radical-Rust:
The money requirements for some of the buildings are fairly high. I felt that the previous values were too low, and that some projects should cost far more than others, especially given their benefits. If you think they're too unbalanced, I am open for suggestions.

settintotrieste:
I like the new stuff and the money requirements are fine balance wise. There are two minor adjustments required though - the french event has an extra 2 brackets in between vassal_of = FRA and war_with = FRA. The scope any_country does not include THIS so you'll need to add the prestige amount for the decision taker.

Radical-Rust:
Fixed the France event. I'm not sure what the other problem is. 
For example, "any_country = { limit = { OR = { owns = 966 owns = 1049 }} prestige = 50 }" gives prestige both to the decision taker, and the nation owning the other territory (if there is one).

Radical-Rust:
So Boolean operators do not work in decisions effect?
I'll fix it in a few hours, tomorrow morning at best.

settintotrieste:
THIS and OR do work however the scope any_country and random_country does not include the country that fires the event or decision. So if you want to apply the effect to every country you’ll need THIS = { prestige = 25 } and any_country = { prestige = 25 }

Radical-Rust:
I had wrongly assumed houtribdijk was part of the Delta Works so I deleted it, thanks for pointing that out.
Everything should be fixed now.

PR #541, opened 2020-03-10
2 new construction projects + small tweaks

Radical-Rust:
- Added Sydney Opera House and Brenner Base Tunnel construction decisions;

- Added 2 new terrorist events, "Chemical attack" and "Separatist attack";

- Tweaked various prestige and money values;

- Other small fixes.


settintotrieste:
I approve of the additions. There are some minor fixes that are needed:
1. There is a ; in the brenner_base_tunnel_desc text. This will cut off the text at the ;.
2. Event 1511126 should give only 10 flashpoint tension when the diplomatic option is taken
3. The tunnel events have missing “=“ in “owns = “
4. brenner_base_tunnel has NOT = { has_global_flag = channel_tunnel } in the trigger. It should be the flag for the brenner tunnel
5. It is not necessary to use THIS = { } for the decision taker
6. Bribe aristocrats should only charge 500 per province. This is because the amounts can quickly spiral into huge ones for big countries like China or Russia with 100s of provinces.
7. 1511126 Should have a check for is_capital = no to avoid 1 province nations from gaining flashpoint tension.

Radical-Rust:
All should be ok now.

settintotrieste:
Thanks. You accidentally put the 500 deduction for the bribe aristocrat decision as a positive number though

Radical-Rust:
Whoops. Fixed it.
I'll try to be more careful in the future.

PR #548, opened 2020-03-18
Fixes scandinavia not inherting scandinavian countries

nichlas0010:
See title.

For whatever reason, only Åland gets inherited when using the event, using annex_to everyone does.


settintotrieste:
Used a better fix in #548

nichlas0010:
what

PR #564, opened 2020-07-07
Update army_tech.txt

TK3600:
Adjust bomber support stat, making it stronger on late game and weaker early. Increase some organization increase from tech. Increased supply consumption as tech increase to simulate the increased cost of warfare, and to balance weapon price as production increase.

### DIFF
--- a/CWE/technologies/army_tech.txt
+++ b/CWE/technologies/army_tech.txt
@@ -675,7 +675,7 @@ modern_rifles = {
 	unciv_military = yes
 
 	army_base = {
-		supply_consumption = 0.02
+		supply_consumption = 0.1
 	}
 
 	ai_chance = {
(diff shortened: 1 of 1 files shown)


settintotrieste:
Supply consumption needs to be kept low because of the hard coded 2000 goods per day limit. If supply consumption is too high, countries with large armies will not be able to get enough supplies to fund it.

TK3600:
If we go by a relatively expensive unit like tank, it can supply 4.44 million man standing army made of entirely tanks.

settintotrieste:
What I mean is that there is a hardcoded limit to how much supply can be purchased per day (2000) hence if the supply limit is too high then countries with large armies would be unable to get fully supplied. There have been some reports of this already happening hence I’ll reject this rather large increase in supply consumption.

TK3600:
Fair enough, I will remove the supply increase part.

PR #565, opened 2020-07-07
Update culture_tech.txt

TK3600:
Increased science's research boost, increased education's education boost. The boost is countered by reduced boosts in inventions. It is changed because satellite should not be 5 times better at teaching than actual education techs.


settintotrieste:
I think the increase is too much.

TK3600:
The increase is countered by decrease of boost in invention part. I think the net effect is neutral no change.

PR #571, opened 2020-07-08
Levantine Federation and Israeli-Palestinian Commonwealth

majorcygnus:
-Added Levantine Federation(Formed via pre 1970 Populist in Israel)
-Added Israeli-Palestinian Commonwealth(Formed by the British or Israeli Communist)
-Added ability to give the whole Mandate of Palestine to Israel
-Updated Israeli Political Parties


settintotrieste:
I will need to do some changes

PR #572, opened 2020-07-08
Update culture_tech.txt

TK3600:
Readjusted earlier research modifier to reduce it. Now it works in a curve to account for exponential increase.

### DIFF
--- a/CWE/technologies/culture_tech.txt
+++ b/CWE/technologies/culture_tech.txt
@@ -53,7 +53,7 @@ information_theory = {
 	cost = 7200
 
 	prestige = 0.05
-	increase_research = 0.05
+	increase_research = 0.075
 
 	ai_chance = {
 		factor = 1.01
@@ -79,7 +79,7 @@ medical_science = {
 	activate_building = health_factory
 	
 	prestige = 0.05
-	increase_research = 0.05
+	increase_research = 0.075
 
 	ai_chance = {
 		factor = 1.01
@@ -103,7 +103,7 @@ big_science = {
 	cost = 12600
 
 	prestige = 0.05
-	increase_research = 0.05
+	increase_research = 0.10
 
 	ai_chance = {
 		factor = 1.01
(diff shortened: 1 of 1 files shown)


settintotrieste:
I feel that the current increase in education efficiency is too high. Literacy is linked to craftsmen promotion so if it is too high then it will affect balance of some states with large populations.

TK3600:
The increased education is countered by nerfing some overpowered invention related to space. In fact it should reduce overall education tech. I too am aiming to nerf high populatio countries like China and India, our vision is the same.

TK3600:
The nerfed invention result in 5 x -15% = -75% education efficiency you can grab rather early, which in my China run resulted me reach American literacy level in 80's. 

Instead I want to shift some of the bonus into tech itself, removing the short cut of rushing detente. 

This should force China to research education tech and neglect their military for a while, much like what China did IRL to develop peacefully.

settintotrieste:
I would advise against having too much education because literacy in game represents university education. China had only around 4% university entrance rate in 1994 which is much less than that of America which I think is mirrored quite closely to in the mod right now.

TK3600:
I told you the invention already reduced education efficiency by 75% in 60's, the increase in education in tech was meant to compensate that. The 75% is far more than 50% increase if you research EVERY education tech. 

At my current testing run, China has 16.7% literacy in 2020 June 14.

www.oecd.org › education › EAG2019_CN_CHN

According to OECD China has 18% with tertiary education completed in 2018. Even with my adjustment, China is not doing better than IRL. But of course, it is AI, a human player would do better, probably meet IRL level just right.

However, if knowing that you prefer even less education efficiency than prior my change, that is fine too. I will reduce the education efficiency.

settintotrieste:
The invention bonus was not intended to be kept. I want to reduce education efficiency across the board. This is to make education reforms more significant in raising literacy.

TK3600:
Alright, I will adjust the value of education efficiency.

PR #581, opened 2020-07-12
Update buildings.txt

TK3600:
1. Increased the material build cost of factory by 10x. This will balance the lack of need for raw material.
2. Increased build time of factory by 2x to slow down unreasonably fast industrialization of developing countries, with a few exceptions. 
> raw materials factory are faster to make, to balance the demand increase and simulate early period industrialization start with low cost goods.
> computer related factory are faster to make to simulate the internet boom.

### DIFF
--- a/CWE/common/buildings.txt
+++ b/CWE/common/buildings.txt
@@ -32,12 +32,12 @@ fish_canned_food_factory = {
 	cost = 0
 	goods_cost = 
 	{
-		electric_gear = 1
-		cement = 1
-		steel = 1
-		machine_parts = 1
+		electric_gear = 10
+		cement = 10
+		steel = 10
+		machine_parts = 10
 	}
-	time = 30
+	time = 60
 	visibility = yes
 	onmap = yes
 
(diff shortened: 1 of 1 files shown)


settintotrieste:
I cannot accept this change.

1. The values must remain at 1 because of many modifiers that increase factory cost. I have attempted to increase costs before only to realise that in certain communist countries, the cost of a factory becomes astronomical.

2. It is necessary for factories to be built quickly to keep up with pop growth. Especially since each level only increases the workforce by 10,000. It is better to balance the growth of craftsmen population since it is the number of employed craftsmen alone that determines industry score.

TK3600:
It seems alright in China. What are the examples of communist country that suffer from it? I will look into it in my playtest.

settintotrieste:
IIRC it was Russia and North Korea.

TK3600:
Alright, I will keep the files on my end until I am confident that Russian and North Korea will work. We can discuss then.

settintotrieste:
Thanks

PR #583, opened 2020-07-12
balancing pop needs and military units.

TK3600:
See commits for details.

### DIFF
--- a/CWE/poptypes/aristocrats.txt
+++ b/CWE/poptypes/aristocrats.txt
@@ -13,7 +13,10 @@ rebel = {
 }
 
 life_needs = {
-	grain = 1.5
+	cattle = 0.75
+	fish =  1
+	fruit =  1
+	canned_food =  2.5
 }
 
 everyday_needs = {
@@ -36,7 +39,8 @@ everyday_needs = {
 	paper = 1
 	telecommunications = 2
 	transport = 2
-	small_arms = 0.5		
+	small_arms = 0.5	
+	computers = 0.2
 }
 
 luxury_needs = {
@@ -52,7 +56,8 @@ luxury_needs = {
 	financial_services = 10
 	clothes_footwear = 2
 	aeroplanes = 2
-	ships = 2
+	ships = 2 
+	computers = 1
 }
 
 country_migration_target = {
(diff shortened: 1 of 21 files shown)


settintotrieste:
I cannot accept this change.

1. Life needs cannot be more than 1 good. This due to a strange bug which prevents the other goods from being bought.

2. I prefer to keep the current list of luxury and Everyday goods. Certain goods like computers tend to be under produced by the AI so it can affect the base tax rate if they are Everyday goods.

3. I think you accidentally reverted the issues tweaks done in the previous version.

4. Supply cost for military units should not be increased. This is because of the 2000 goods per day purchase limit and the large number of brigades in the late game which can cause many brigades to remain unsupplied.

TK3600:
How do you like the idea of increase military goods value by 5 or 10, but decrease the output by the same?

settintotrieste:
I prefer to keep the output the same because of the sphere duplication bug that causes goods produced in spheres to be unobtainable by countries outside the sphere but yet is still reflected in the global supply.

TK3600:
IIRC if there are any left over in the sphere, it is given to global market.

TK3600:
Anyway, the idea is to make the military goods to maintain the same profitability, but increase their cost to make poor countries unit spamming harder.

settintotrieste:
It does not always get sent to the global market. There’s a guy on reddit who discovered this bug and I’ve witnessed it before as sphered Peru where ships were not able to be built despite ample supply in the global market. It is because of this I prefer to make sure there’s more military goods in the market than usual.

TK3600:
Technically if the supply consumption is halved, there should be no net difference.

settintotrieste:
You can explore that option.

TK3600:
Alright, testing it.

settintotrieste:
Thanks

TK3600:
OK, there is actually a side effect. Raising price 2x led to greater price ceiling increase more than merely double. This led to an early starvation of poor country not able to afford heavier unit until 60's. Realistic? Yes. But some players wanting to play African minor might be upset. I could lower the cost of military goods a little, if you think so.

settintotrieste:
I’m not too concerned with realism for the military situation since that’s not possible to accomplish in Vicky so the next best thing is fun. This is going to be a big problem for fun since they’re quite a few wars in the 40s and 60s involving minor powers. I think the prices should remain until a solution is found that does not affect weaker countries having sufficient units.

TK3600:
OK I lowered the price a bit, and reduce consumption of light infantry and irregulars. The problem is now solved. Would you like to see the changes?

PR #584, opened 2020-07-14
2000 Supply fix

TK3600:
Same as last time, except with supply adjustment. Supply halved, cost of supply doubled, net effect no change, but get around the 2000 supply bug. I adjusted the price a bit, now minors should able to afford army just fine.

You may want to edit the poptype part where the reform issue was not included.

### DIFF
--- a/CWE/common/goods.txt
+++ b/CWE/common/goods.txt
@@ -1,13 +1,13 @@
 military_goods = {
 
 	small_arms = {
-		cost = 2.4 
+		cost = 4.8 
 		color = { 208 202 127 }
 		overseas_penalty = yes
 	}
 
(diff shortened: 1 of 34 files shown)


settintotrieste:
I've decided to add a new military good - ammunition. It would take over as the main supply good for military units. Small and Heavy arms will continue to be consumed by soldier and officer pop just to keep demand there all the time but other than that, those goods will be construction goods for units.

TK3600:
That is an interesting approach. While I do think my system is actually very stable after fix, I am curious of how the ammo system works. As far as weaponry goods goes, it is not consumed by any pop, so its consumption would become entire based on construction of units.

PR #609, opened 2020-07-23
Update Multi-Role Fighter.txt

TK3600:
Extremely expensive late game unit that can fulfill both the role of backline and frontline.


settintotrieste:
I prefer not to use aero planes and optics as supply goods since those aren’t parts that regularly get changed.

PR #613, opened 2020-07-25
Update aristocrats.txt

TK3600:
Executives are too used to urban life to eat unprocessed grain.


settintotrieste:
I cannot accept this change because I set life needs to be the same for all pops to represent the barest minimum food a person, whether rich or not, needs to live.

PR #614, opened 2020-07-25
Update capitalists.txt

TK3600:
Investers refuse to eat unprocessed grain over processed food.


settintotrieste:
I cannot accept this change because I set life needs to be the same for all pops to represent the barest minimum food a person, whether rich or not, needs to live.

PR #644, opened 2020-09-15
Update Essential events 3.txt

TK3600:
large undeveloped democracy is not immune to massive societal corruption anymore, like IRL.


settintotrieste:
I have to reject this for balance reasons because India has the licence Raj. They can’t get too many maluses or their factories will produce nothing.

TK3600:
I have submited a pull request with adjusted license raj modifier. Factory malus and other redundant penalty are taken out. See if this solve the OP india problem at last.

PR #654, opened 2020-09-20
Swiss culture tweaks

Outta-my-way-burrocrat:
The Romansh are Swiss not Romanian


settintotrieste:
I’m afraid there is some corruption in the cultures.txt file causing all the accented characters are messed up. Please use ANSI encoding for the cultures.txt file.

PR #685, opened 2020-10-02
Update buildings (for reference).txt

TK3600:
These are my changes. You dont have to take them in, I am just sending this in case you want to replicate it.

### DIFF
--- a/CWE/common/buildings.txt
+++ b/CWE/common/buildings.txt
@@ -35,7 +35,7 @@ fish_canned_food_factory = {
 		electric_gear = 20
 		cement = 20
 		steel = 20
-		machine_parts = 10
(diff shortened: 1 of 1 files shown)


TK3600:
Note the increase in cost is 20x.

settintotrieste:
Maybe the goods cost can be increased by only 5 times? I know that the numerous modifiers in this mod can make a base cost of 1 go up to 5-7 so if its goods cost = 5 then it'll likely go up to 25-35 in game.

settintotrieste:
Would it be possible if you could help me do one last test with the GP rankings for the updated costs?  I just want to double check the GP and SP ranking in 1985 before integration.

PR #686, opened 2020-10-02
Update production_types.txt

TK3600:
These are the changes to accommodate increased demands for base building materials. You dont have to take them in. I am posting here for a reference.

### DIFF
--- a/CWE/common/production_types.txt
+++ b/CWE/common/production_types.txt
@@ -367,7 +367,7 @@ oil_seeds_canned_food_factory = {
 cement_factory = {       
 	template = factory_template
 	input_goods = {           
-		coal = 5
+		coal = 6
 	} 
 	output_goods = cement 
 	value = 10
--- a/CWE/common/production_types.txt
+++ b/CWE/common/production_types.txt
@@ -387,7 +387,7 @@ cement_factory = {
 steel_factory = {       
 	template = factory_template
 	input_goods = {    
-		iron = 6	
+		iron = 7	
 		coal = 5	
 	}
 	output_goods = steel 
--- a/CWE/common/production_types.txt
+++ b/CWE/common/production_types.txt
@@ -387,11 +387,11 @@ cement_factory = {
 steel_factory = {       
 	template = factory_template
 	input_goods = {    
-		iron = 8	
-		coal = 5	
+		iron = 12	
+		coal = 6	
 	}
 	output_goods = steel 
-	value = 6.3
+	value = 8
 	bonus = {
 		trigger = { 
 			country = { has_country_flag = spclow }
--- a/CWE/common/production_types.txt
+++ b/CWE/common/production_types.txt
@@ -660,10 +660,10 @@ electric_gear_factory = {
 bio_fuel_factory = {       
 	template = fuel_template
 	input_goods = {    
-		grain = 4
+		grain = 4.5
 	} 
 	output_goods = fuel 
-	value = 3
+	value = 2
 	bonus = {
 		trigger = { 
 			country = { has_country_flag = spclow }
@@ -719,7 +719,7 @@ gas_fuel_factory = {
 coal_fuel_factory = {    
 	template = fuel_template
 	input_goods = {    
-		coal = 3
+		coal = 4
 	} 
 	output_goods = fuel 
 	value = 2.2
--- a/CWE/common/production_types.txt
+++ b/CWE/common/production_types.txt
@@ -545,8 +545,8 @@ machine_parts_factory = {
 aluminium_factory = {       
 	template = factory_template
 	input_goods = {    
-		bauxite = 4.5
-		coal = 2.5
+		bauxite = 6
+		coal = 3
 	}
 	output_goods = aluminium 
 	value = 6
--- a/CWE/common/production_types.txt
+++ b/CWE/common/production_types.txt
@@ -387,11 +387,11 @@ cement_factory = {
 steel_factory = {       
 	template = factory_template
 	input_goods = {    
-		iron = 12	
+		iron = 16	
 		coal = 6	
 	}
 	output_goods = steel 
-	value = 8
+	value = 10
 	bonus = {
 		trigger = { 
 			country = { has_country_flag = spclow }
--- a/CWE/common/production_types.txt
+++ b/CWE/common/production_types.txt
@@ -367,7 +367,7 @@ oil_seeds_canned_food_factory = {
 cement_factory = {       
 	template = factory_template
 	input_goods = {           
-		coal = 6
+		coal = 7
 	} 
 	output_goods = cement 
 	value = 10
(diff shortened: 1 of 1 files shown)


settintotrieste:
One of the problems I noticed is that military scores are so huge that China's total score will exceed the USSR by the 80s even with the current soldier pop debuff, which is going to be more pronounced because industry scores are greatly reduced here. I think military unit supply_consumption has to be reduced.

TK3600:
I think the opposite. More consumption increase by tech will make advanced military weight higher in supply vs sheer number.

Solution: Decrease base military consumption, increase military consumption via tech. Keep status quo overall.

TK3600:
Give me some time, I will submit an updated unit supply stats and tech for review.

settintotrieste:
I believe that all economic or military updates should follow these general principles:
1. Supply consumption should be reduced overall because of the 2K daily goods limit cap which is a problem for countries have 2K brigades and excessively high military scores will mess with the current GP list.
2. Factories base production should be as low as possible to reduce late game lag because even a +1 increase can scale up massively when factory levels rise.
3. Consumption should not be so high or production so low that there is an acute shortage of goods among the GPs to build factories except for the first 5 years.

TK3600:
I too firmly believe that. However there are way to achieve intended balance with few tricks.

1. Knowing the 2000 good cap, we can make unit consume many different product so not one is being used too much, yet still have high overall cost.

2. What you did, add extra consumables like ammo.

3. My old proposal, increase cost of goods 2x, halve the production rate.

As of now, I think the current balance is good. I am open to opinion on what to change though.

settintotrieste:
Regarding the 3rd point, the base profitability should be around 25% so that affects the margins quite a bit. Using modifiers to compensate runs the risk of the stack overflow problem.

TK3600:
Now that unit consumption has decreased, would this change be approved?

settintotrieste:
I’ve been thinking over the last few days whether it is worth it to increase consumption. An increase in factory goods cost by x200 will further increase goods production on top of the already slow creep in production output over the last few months. That’s going to further worsen the noticeable lag starting from the 80s. This is also going to raise the spectre of the stack overflow problem since more goods sold = more money in the system. I think the better approach is to change cost = 0 to something like cost = 5. That provides a better limit to industrialisation since the national bank is dependent on factory profitability or government investment which doesn’t rely on fickle sphere bugs or prestige goods purchase rankings. This also solves the issue with national banks getting past the stack overflow limit since cost vanishes into thin air.

TK3600:
Trust me it works. Plenty of material for minors to build. Can you merge it and see how players think?

TK3600:
5 gold per factory will not make any difference. That is really a whole other can of worm. Though i can work on it in mean time.

TK3600:
>  An increase in factory goods cost by x200 will further increase goods production on top of the already slow creep in production output over the last few months. That’s going to further worsen the noticeable lag starting from the 80s. 

Hold on, I dont quite get it. If anything factory cost 200x more steel and cement to make should make production reduced. The lag should decrease.

settintotrieste:
I won't merge it unless I'm sure it's good. I've had a lot of bad experiences merging stuff without first checking them thoroughly. This mod is like a tower of cards so we gotta be careful with big changes. 

Sorry for not being clear, what meant is if the total number of factories is reduced then the production rate of each factory has to be increased to meet existing demand. Because this is caused by building costs, once those construction goods have reached critical mass and no longer pose a problem, factory numbers start to rise quickly (you'll see this in NWO), excess production will then appear and increase significantly (this is possible because the mod gives free money to stop rolling bankruptcies like in vanilla). This was major a problem previously where there was an acute shortage in the early game so I increased base factory output just increased exponentially over time. It was not a problem until the 1980s but by 2010, factory input requirements exceeded RGO output so many countries had factories that produced nothing.

So these are the 3 problems that goes against accepting this change:

1. There is already a goods shortage. People are saying that they can't fulfill pop needs well. Reducing factories will make this problem worse. The problem is now how to address this without causing the problem above?

2. Less factories means less craftsmen, that diminishes their political participation in the mod. I'd like to keep them there to simulate modern industrialised politics. 

3. Less industry score means I have to look at the other scores again such as prestige. If there's too much prestige, certain countries remain where they are for too long. I'm not a fan of chain changes because it is really tedious redoing old stuff.

TK3600:
1. This problem only happens in first 5 years. The shortage is no longer a problem as in my report after the production type adjustment. In fact, I would attribute this as part of another problem: the starting date have all country underindustrialized, so you see a massive catch up happening.

2. This is indeed a problem, though I dont think this change will affect it significantly as factory while reduced not by THAT much. Though if we do increase the initial factory amount in start date, this will go away too.

3. This one depends on if your military consumption change will suffice, along with our soldier pop change. I have not tested so I leave no comment.

settintotrieste:
Okay, but can you decrease the construction cost to only 5 max? Because based on your test run to 1976, there are still some noticeable problems. First is Australia became a GP which I never got in normal test runs, secondly France is not a GP which is worrying if they weren't dismantled. Thirdly, the industry score of the USSR is much lower than previously, I want the USSR to have at least half of the US's score as much as possible. 

And what is the actual goods cost in the USSR in 1976 after all the modifiers? Is it more than 50 per construction good?

TK3600:
I looked into it. France is normal, they have super low prestige due to losing colonial wars.

Soviet Union is odd. In practice the cost of building is 37 steel, which is not that bad in itself. However, Soviet AI has a whole bunch of factory queued and unable to build due to lack of steel. Literally half their industry is using steel, they spam a shit ton of car factory and ship factory. They dont have enough steel to build more factory. The investors oddly are able to build and get required steel.

TK3600:
On top of that, while investors can build and get their steel, they take like 270 steel to bulid anything.

TK3600:
So it turns out this is a much bigger economy problem that the original creator failed to address and hid it with non-existing factory cost. Now we are cleaning after their mess. Lucky for you, I am someone obsessive enough to fix it. I will report further with the steel buff, the throughput should be doubled to meet the demand. Incidentally the global cheapo price of iron and its problem causing labourers to starve will also be addressed by steel eating more iron.

settintotrieste:
Instead of escalating steel production, wouldn't it be better to just reduce the construction cost for steel?

TK3600:
The thing is prestige really do matters now that mil score is fixed, and early industry score is low. Often you see prestige is more than industry and military combined.

As for steel production, it is actually very well balanced atm.

Goods do not cause lag, factory causes lag.

settintotrieste:
Either way, you’ll need to address that massive infrastructure cost problem for the USSR first

TK3600:
First I need to find the modifier that causes it, which I suspect is factory owner cost.

Second I need to know what event gives those modifier, and I need to change them into a sensible level.

Then we can address the problem and begin to implement these changes.

Frankly our attention should no longer be about the factory resource cost. It just revealed a bigger problem masked by non-existent factory cost.

settintotrieste:
Factory owner cost is affected by country or province or triggered modifiers. There are 100 modifiers that affect factory_owner_cost in event_modifiers.txt and 11 in triggered_modifers.txt

settintotrieste:
With the reduction in factory costs overall, is the increased steel and lumber production still required?

TK3600:
I think by itself this pull request is already balanced.

settintotrieste:
Okay, I integrate this

PR #687, opened 2020-10-04
Update army_tech.txt (addressing china soviet union mil score disparity)

TK3600:
Adjusted supply consumption with consideration of military score of developed nation vs developing nation. Tank tech now affect supply of affect unit type rather than all units.

### DIFF
--- a/CWE/technologies/army_tech.txt
+++ b/CWE/technologies/army_tech.txt
@@ -2070,7 +2070,7 @@ proxy_war = {
 	regular_experience_level = 5
 
 	army_base = {
-		supply_consumption = 0.10
+		supply_consumption = 0.05
 	}
 
 	ai_chance = {
@@ -2149,7 +2149,7 @@ assymetric_warfare = {
 	regular_experience_level = 5
 
 	army_base = {
-		supply_consumption = 0.10
+		supply_consumption = 0.15
 	}
 
 	ai_chance = {
@@ -2182,7 +2182,9 @@ arms_race_doctrine = {
 	morale = 0.25
 	unciv_military = yes
 	military_tactics = 0.25
-
+	army_base = {
+		supply_consumption = 0.20
+	}
 	combat_width = 1
 
 	supply_limit = 0.1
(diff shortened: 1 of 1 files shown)


settintotrieste:
I think what should be done instead would be to remove all army_base supply costs increase for all techs other than those from the army and navy doctrine branch and use individual unit increases instead. Also supply_consumption should only be increased by 0.05 each time.

TK3600:
The consumption should be OK, I lowered base cost of units, it should add up to status quo. I think the number is reasonable if you see how much larger vanilla vic 2 consumption increases.

If you lower the cost, we run into problem of india and china have too much mil score for low tech army.

If you lower base cost only, China will build more army now they have more money, manpower is never an issue for them.

settintotrieste:
The problem is that the amount of supply consumed actually increases military score.

TK3600:
Mil score is only an early game problem. Industry become the main source of score in mid and late game. By reducing base consumption, it will matigate the early game problem. The tech increase should not actually impact ranking, if not helping developed country have better mil score than lesser ones.

TK3600:
I define early game as in 1976 and earlier. If you check out my latest report, the ranking is actually reasonable.

settintotrieste:
From my test run with the current industry set up and no major wars but just before the -50% supply cost reduction tweak, the PRC had 16K military score which made them rank 2 displacing the USSR in 1981. China had only 169 industry score and 2K prestige compared to Russia which had 3.6k industry score, 4K military score and 6K prestige. After applying the -50% supply reduction modifier to the savegame, China’s score shrank to 7.8K while Russia shrank to 3K. China was still no.2.

TK3600:
How did China afford so much army?

TK3600:
I honestly think it is a separate problem the fact someone can afford so much army....

And why is Russian prestige so low? Soviet collapsed?

settintotrieste:
The AI will always build units to max out their cap until it goes bankrupt then they will disband all units. Even with 5% soldier pop, China will still have around 500-700 divisions and the supply consumption of that jacks it up to astronomical levels. Personally I don’t find having a large army a problem because due to the code, it is either a large army or a small one for all nations. I prefer the former because it helps the AI to finish off the historical wars without taking forever and gives players of smaller nations a chance to do something fun.

PR #691, opened 2020-10-06
Update farmers.txt (please check syntax)

TK3600:
Attempt to reduce soldier pop % if country is large: >200mil population.

### DIFF
--- a/CWE/poptypes/farmers.txt
+++ b/CWE/poptypes/farmers.txt
@@ -1182,16 +1182,22 @@ modifier = { factor = 5 country = { has_country_flag = indo_con } }
 				soldiers = 0.035
 			}
 			modifier = {
-				factor = -2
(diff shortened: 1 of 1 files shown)


settintotrieste:
Okay I forgot that I did some changes to the soldier limit during the last major update by importing HPM's variables. The absolute limit to soldier percentage is 5%. I think that is reasonable regardless of how populous a country is. I have however removed the buff (max +4% to soldier promotion) through conscription laws and made a hard limit to soldier pops based on conscription reforms when there is no war so for example if 3 year draft is not selected, the max soldier pop is 4%. This should slow down promotion rate so China's military score should be less in the 80s.

TK3600:
OK, lets leave this for now and see how it goes.

settintotrieste:
Okay, I'll close this PR for now.

PR #693, opened 2020-10-07
Update goods.txt

TK3600:
Doubling cost of military goods, while halving the consumption = mil score halved!


TK3600:
Cost of military goods doubles, while  we halve their consumption. We managed to halve mil score without touching their actual upkeep price!

settintotrieste:
That's great!

PR #706, opened 2020-10-08
Update Nuclear Aircraft Carrier.txt

TK3600:
mil score rebalance


settintotrieste:
I'll keep nuclear aircraft carriers expensive

TK3600:
Believe me it still is. It cost exactly the same as before. The cost of ammo doubled, so halving it makes it the same as before. Though regardless of ammo cost ammo is like only 10% of its consumption cost, so by stopping this there would be little difference.

PR #716, opened 2020-10-08
fixing soviet running out of steel bug.

TK3600:
Trying to help AI production here. Due to numerous owner cost modifier, it is better not to invest in rail projects rather build it directly.


settintotrieste:
I suppose there is no need for this now that the modifiers have been adjusted

TK3600:
Honestly I am still in the middle of editing. I didnt fully test it yet, but soviet does seem to do better by 1950.

PR #721, opened 2020-10-09
Update army_tech.txt

TK3600:
Now that mil score is fixed, we can finally implement this. I reduced a bit from previously one just in case. The increase are mainly in late game, where industry should have rised quite a bit. This will make sure mil score wont be too high.

please please take it in, lets not fight over it.

### DIFF
--- a/CWE/technologies/army_tech.txt
+++ b/CWE/technologies/army_tech.txt
@@ -60,7 +60,7 @@ strategic_bomber = {
 	}
 	
 	army_base = {
-		supply_consumption = 0.10
+		supply_consumption = 0.05
 	}
 
 	dig_in_cap = 1
@@ -109,7 +109,7 @@ icbm = {
 	unciv_military = yes
 
 	army_base = {
-		supply_consumption = 0.10
+		supply_consumption = 0.05
 	}
 
 	max_fort = 1
@@ -2110,7 +2110,7 @@ mutual_assured_destruction = {
 	regular_experience_level = 5
 
 	army_base = {
-		supply_consumption = 0.10
+		supply_consumption = 0.05
 	}
 
 	ai_chance = {
@@ -2149,7 +2149,7 @@ assymetric_warfare = {
 	regular_experience_level = 5
 
 	army_base = {
-		supply_consumption = 0.15
+		supply_consumption = 0.10
 	}
 
 	ai_chance = {
@@ -2183,7 +2183,7 @@ arms_race_doctrine = {
 	unciv_military = yes
 	military_tactics = 0.25
 	army_base = {
-		supply_consumption = 0.20
+		supply_consumption = 0.15
 	}
 	combat_width = 1
 
@@ -2289,7 +2289,7 @@ network_centric_warfare = {
 	regular_experience_level = 5
 
 	army_base = {
-	supply_consumption = 0.25
+	supply_consumption = 0.20
 	}
 	
 	ai_chance = {
--- a/CWE/technologies/army_tech.txt
+++ b/CWE/technologies/army_tech.txt
@@ -1,11 +1,14 @@
-#increase fort, dig-in cap, decrease attrition, increase army base supply consumption
+#ballistic_and_recon increase seige of bombers and manuv of fig
 post_war_radar = {
 	area = ballistic_and_recon
 	year = 1946
 	cost = 3600
 	unciv_military = yes
 
-	activate_building = fort
+	activate_building = fort	
+	army_base = {
+		supply_consumption = 0.05
+	}
 
 	dig_in_cap = 1
 	
@@ -57,7 +60,7 @@ strategic_bomber = {
 	}
 	
 	army_base = {
-		supply_consumption = 0.05
+		supply_consumption = 0.10
 	}
 
 	dig_in_cap = 1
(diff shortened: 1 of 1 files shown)


settintotrieste:
I won't be accepting this because it once again dramatically increases supply_consumption. The max increase possible per level should only be 5% unless it is multi_role.

TK3600:
Believe it is actually very balanced, I tested it.

TK3600:
Why dont you let it through. I have a good track record. Further more I have proof that in vanilla it works.

TK3600:
In vanilla tech make unit cost 10% more minimum, not even counting inventions. The cost of maintaining military is a huge component of win and loss of cold war.

TK3600:
point_defense_system = {
	area = army_doctrine
	year = 1850
	cost = 10800
	unciv_military = yes
	engineer = {
		support = 0.25
	} 
	army_base = {
		supply_consumption = 0.20
	} 
	max_fort = 1
	dig_in_cap = 1
	ai_chance = {
		factor = 1.01
		modifier = {
			factor = 1.2
			civilized = yes
			any_neighbor_country = {
				civilized = yes
				NOT = { is_greater_power = yes }
				military_score = THIS
			}
		}
		modifier = {
			factor = 1.05
			war = yes
		}
		modifier = {
			factor = 1.1
			year = 1860
		}
		modifier = {
			factor = 2
			year = 1870
		}
		modifier = {
			factor = 1.4
			is_greater_power = yes
			any_neighbor_country = {
				is_greater_power = yes
				military_score = THIS
			}
		}
		modifier = {
			factor = 1.02
			tech_school = army_tech_school
		}
		modifier = {
			factor = 1.1
			is_greater_power = yes
		}
	}

}
deep_defense_system = {
	area = army_doctrine
	year = 1870
	cost = 14400
	unciv_military = yes
	engineer = {
		support = 0.25
	} 
	army_base = {
		supply_consumption = 0.20
	} 
	max_fort = 1
	dig_in_cap = 1
	
	ai_chance = {
		factor = 1.01
		modifier = {
			factor = 1.2
			civilized = yes
			any_neighbor_country = {
				civilized = yes
				NOT = { is_greater_power = yes }
				military_score = THIS
			}
		}
		modifier = {
			factor = 1.05
			war = yes
		}
		modifier = {
			factor = 1.1
			year = 1875
		}
		modifier = {
			factor = 2
			year = 1880
		}
		modifier = {
			factor = 1.4
			is_greater_power = yes
			any_neighbor_country = {
				is_greater_power = yes
				military_score = THIS
			}
		}
		modifier = {
			factor = 1.1
			is_greater_power = yes
		}
		modifier = {
			factor = 1.02
			tech_school = army_tech_school
		}
	}
}

The above is the vanilla example.

I mean come on, you added like 100 modifiers in the range of 20% or more, and I cant be trusted to elevate things to vanilla level? I am not sure whether to be flattered or frustrated that I am being held at an impossible standard.

You requested me to fix mil score first, so I did. Then you said doing so affected overall ranking, so I fixed factory and economy. Now I finally finally able to perfectly correct things, I am getting blocked. :/

settintotrieste:
I understand you really want to add in this but I’m afraid I don’t want to. Supply consumption shouldn’t be increased through techs like vanilla because this mod is way different from vanilla. For one, basic consumption is already much higher than vanilla. Light infantry costs more goods than Guards. There’s also the issue with massive soldier caps because of the massive global population. China can build a 1000+ units and mobilise another 1000+. There are also numerous modifiers that boost supply consumption without techs which you don’t get in vanilla. Military score as it stands now is also good enough. In one of my test games, the UK had about 750 military score when industry score was only 500 in 1958. I think it is a sign that consumption shouldn’t be dramatically raised when a simple 20% increase in consumption through the communist bloc leader modifier can make the USSR’s score jump from 700 something to 1700+ in the 1950 start.

TK3600:
I am very much in agreement with you that mil score should not be key determinant of GP ranking. I assure you by 1980's industry score should not be behind mil score. And China would not overtake USSR before 1980, and usually not until USSR collapse in 1990's.

In fact, I have put lots of thought and detail into the change. I specifically made sure that early game is 5% increase, or none at all. Only in later game by 1960+ does it start to get a bit ramped up. You can say pre-1970 consumption is less or equal to yours.

In fact in one way or another, I found the unit money cost actually dropped by 2 times before this whole thing started. I want it to raise back to that level by 1980's.

As for consumption vs vanilla, this mod also has much increased production bonus and much much MUCH more money floating around.

TK3600:
So in conclusion while it is true around 1960's mil score is higher than industry score. It will be balanced out in 1970. I had game running in background multiple times well into 21 century. I really put time into my tests.

Reason being industry score grows exponentially, as industry grow more profitable more will be built. So would the population increase with literacy. 

Mil score is more or less linear with tech being the only increase with tech (5% at time) and population.

settintotrieste:
The GP AIs are supposed to have unlimited money in this mod so I'm not concerned about that. I did a test run yesterday from 1946 to 1960 and China's score jumped to no.1 by 60 with about 3K military score. The US was somewhere around 2.7K. I doubt China took many military techs so there's already there. I'll compromise with you and add around 10% supply consumption per level to the military doctrine branch. That is the max I will do.

TK3600:
OK, that is fair.

PR #723, opened 2020-10-09
Update culture_inventions.txt

TK3600:
Nerfing the busted darwinism education boost, while redistributing it to tech.


settintotrieste:
I'm reducing the tech's education boost to 10%, that is the max any invention or tech should give at a time

PR #725, opened 2020-10-09
Update tank.txt

TK3600:
Removed canned food to avoid beefing up mil score too much.


settintotrieste:
Canned food should remain since that is to represent the soldier upkeep on the field

PR #727, opened 2020-10-09
Sphere bug fix (the forbidden update)

TK3600:
For reference only, just leave it here. If anyone want a taste of this game having sphere bug fixed, try this simple change.

Settin, you know you want to try deep inside!


settintotrieste:
It's interesting but it is better if you don't use Nauru for this submod because they have events and decisions for them

TK3600:
What do you think about Jan Mayen? In game it has 32k population, but irl it has no permanent residence, the army stationed at most was 32. You may want to look into that.

settintotrieste:
If you can create a province for Jan Mayen then maybe I can put this in as an optional extra like performance mode.

TK3600:
That is a brilliant idea, best of both world. I will get started on the work on it.

PR #736, opened 2020-10-14
Elia

Eeillios:
Added all Bantustan related events + Seychelle events tied to South Africa and they are fully working


settintotrieste:
Thanks, I'm getting an error on saying that the pull request cannot be merged. I suspect it is because you did not use the most updated seychelles.txt and south_africa.txt as the base for your changes. For seychelles.txt, those events have already been added last month so all you need to do is to reupload the master branch's file. For south_africa.txt, you would need to re-upload the file using the current master branch version as the base for your changes.

settintotrieste:
It seems like the PR is corrupted by the use of the old south_africa and seychelles files to add changes. It messed up the git directory on your side such that it is not in sync with the master's, causing the error. I think what you need to do is to first do a fetch from the master on your desktop git client. That should not affect any of your changes other than those in the south africa files. After that, you'll need to reapply the changes to the south_africa.txt file and then finally create a PR again.

PR #747, opened 2020-10-31
Update buildings.txt

TK3600:
Forgot to add back money cost to buildings while working on rebalance.

### DIFF
--- a/CWE/common/buildings.txt
+++ b/CWE/common/buildings.txt
@@ -29,7 +29,7 @@ fish_canned_food_factory = {
 	on_completion = factory
 	completion_size = 0.2
 	max_level = 584
-	cost = 0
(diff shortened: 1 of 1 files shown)


settintotrieste:
I've decided to drop the money cost requirement because the costs were rising into the millions even with the factory cost modifiers being fixed. I'll think of a better method to address the cash stack overflow problem.

TK3600:
wait, millions? Which countries?

TK3600:
Soviet Union the worst offender is doing fine in my games.

settintotrieste:
It was China and Vietnam iirc

TK3600:
regardless of this, country paying millions means something is broken and I need to fix it.

PR #754, opened 2020-11-23
Elia

Eeillios:
SKA and the events related should be fully functioning tested the Congo Crisis a few times, haven't tested the decision to rename tho but I don't think anything is broke there. Unrelated note I was wondering if you could add an event to release the other Namibian bantustans if Namibia doesn't get independence when it should and Apartheid survives, didn't do it myself because I was afraid to conflict with the other SAF code


settintotrieste:
When do you think is a good time to release the other bantustans?

Eeillios:
The event should be tied to Namibian independence, when it's definetely avoided so early 90's they'd be released

PR #762, opened 2020-12-08
Fixed coring bug for European Union

DoktorKure:
Fixed a bug in the Pan-European identity (claim greater EU) decision that wouldn't give cores at all.


settintotrieste:
Incorporated in 7be4bc1

PR #770, opened 2021-01-05
Elia

Eeillios:
New flagtype fully working, game doesn't crush and shows up in game, plus the first adjustment for it. Hope you accept the pull request now so that it doesn't mess with any push that may be done before I get other events in


settintotrieste:
Just to check, the socialist and communist flags for north Vietnam and east Austria in the 1950 start are the same right?

PR #778, opened 2021-01-09
Elia

Eeillios:


settintotrieste:
Thanks, there are some thing that I have to reject though. 

Please try to avoid adding long names to small countries. Just Kampuchea is better since the rest are just formal names using Kampuchea.

Please also avoid adding events that end a chain by firing independently like 8501108. It’s better to script the event to be fired directly by the previous event. That set up cause innumerable issues with NWO and was what I had to remove before 1.12.

PR #784, opened 2021-02-11
Added TRE Flavour

Eeillios:
Parties, Flags and events to include Trieste as 7th Yugoslav Republic as OTL Plans + small tweak to Italian Parties


Outta-my-way-burrocrat:
Could you fix the Italian progressive party trade policies too?

PR #786, opened 2021-02-28
Update CHL - Chile.txt

settintotrieste:
Adding technologies to 1945 start (Chile starts in complete disadventage compared to its South American neighbors)


Outta-my-way-burrocrat:
Thanks @edigpp !

PR #787, opened 2021-02-28
Turkish Political Parties Overhaul

erdemug:
- Modified parties and ideologies
- Changed Upper House distribution
- Changed initial Political and Social Reforms


settintotrieste (changes requested):
Thanks, I like these changes. Would it be possible if you could help shorten some of the party names to below 30 characters (including spaces)? This is to prevent the name spilling out of its box and blocking the country score section of the UI. You can use party acronyms if the name can’t be shortened naturally.

erdemug:
- Shortened party names
- Converted Turkish characters

Outta-my-way-burrocrat:
@erdemug you forgot to update the 1992 ruling party

PR #816, opened 2021-08-08
Update CHI - China.txt

Josh-Drake:
Attempted to address slow literacy growth rate issues


settintotrieste:
I'm afraid I have to reject this as the higher education (literacy) level is too high. It is ahistorical and would trigger the pop growth decline system. It is better to increase literacy by increasing research point generation instead,

PR #830, opened 2021-11-17
Italy 50's

Eeillios:
Added 1953 and 1958 elections and events related to them, San Marino Civil War and eventual showoff with America. The Italian Economic miracle decision was turned into an event reliant on the reconstruction decisions and event 8007751 was reworked to be more accurate. Issues persist with a few events, 8020313, 8020318 and 8020319 crash the game. 8007751, 8020314, 8020320, 8020321, 8020322 don't exist.


CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		any_pop = { 
+			limit = { strata = poor }
+			ideology = { factor = 0.03 value = communist }
+		}
+		any_pop = { 
+			limit = { OR = { strata = poor strata = middle } }
+			ideology = { factor = 0.03 value = socialist }
+		}
+		set_country_flag = 1953_ITA_Centrism
+		ai_chance = { factor = 0.7 }
```
settintotrieste, line 512: Please follow the convention of putting the historical option as the first option in all events and if it is not a historical option, the option with the highest ai_chance should be first. In an event with 3 options, the historical option must have at least 80% ai_chance

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		remove_country_modifier = visionary_thinker
+		remove_country_modifier = competent_placeholder
+		remove_country_modifier = imperious_autocrat
+		remove_country_modifier = prince_of_terror
+		remove_country_modifier = efficient_sociopath
+		remove_country_modifier = pig_headed_isolationist
+		remove_country_modifier = ambitious_union_boss
+		remove_country_modifier = dour_economist
+		
+		add_country_modifier = administrative_genius
```
settintotrieste, line 584: This is wrong. It should be add_country_modifier = { name = administrative_genius duration = -1 }

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		remove_country_modifier = visionary_thinker
+		remove_country_modifier = competent_placeholder
+		remove_country_modifier = imperious_autocrat
+		remove_country_modifier = prince_of_terror
+		remove_country_modifier = efficient_sociopath
+		remove_country_modifier = pig_headed_isolationist
+		remove_country_modifier = ambitious_union_boss
+		remove_country_modifier = dour_economist
+		
+		add_country_modifier = backroom_boy
```
settintotrieste, line 623: This is wrong. Please see above.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		remove_country_modifier = visionary_thinker
+		remove_country_modifier = competent_placeholder
+		remove_country_modifier = imperious_autocrat
+		remove_country_modifier = prince_of_terror
+		remove_country_modifier = efficient_sociopath
+		remove_country_modifier = pig_headed_isolationist
+		remove_country_modifier = ambitious_union_boss
+		remove_country_modifier = dour_economist
+		
+		add_country_modifier = great_reformer
```
settintotrieste, line 662: This is wrong. Please see above.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		remove_country_modifier = visionary_thinker
+		remove_country_modifier = competent_placeholder
+		remove_country_modifier = imperious_autocrat
+		remove_country_modifier = prince_of_terror
+		remove_country_modifier = efficient_sociopath
+		remove_country_modifier = pig_headed_isolationist
+		remove_country_modifier = ambitious_union_boss
+		remove_country_modifier = dour_economist
+		
+		add_country_modifier = landed_gentleman
```
settintotrieste, line 706: This is wrong, please see above.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		random_country = {
+                       limit = { tag = EGY government = presidential_dictatorship is_vassal = no }
+                       relation = { who = THIS value = 50 }
+        	}
+		random_country = {
+                       limit = { tag = JOR OR = { government = hms_government government = absolute_monarchy } }
+                       relation = { who = THIS value = 50 }
+        	}
+		random_country = {
+                       limit = { capital = 1731 OR = { government = hms_government government = absolute_monarchy } } #Lybia
```
settintotrieste, line 786: Please change capital = 1731 to tag = TRI

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		random_country = {
+                       limit = { tag = PER OR = { government = hms_government government = absolute_monarchy } is_vassal = no }
+                       relation = { who = THIS value = 50 }
+        	}
+		random_country = {
+                       limit = { tag = EGY government = presidential_dictatorship is_vassal = no }
+                       relation = { who = THIS value = 50 }
+        	}
+		random_country = {
+                       limit = { tag = JOR OR = { government = hms_government government = absolute_monarchy } }
```
settintotrieste, line 782: Please insert a check for is_vassal = no.

CWE/events/italy.txt
settintotrieste, line 786: Please insert a check for is_vassal = no.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+                       relation = { who = THIS value = 50 }
+        	}
+		random_country = {
+                       limit = { capital = 1731 OR = { government = hms_government government = absolute_monarchy } } #Lybia
+                       relation = { who = THIS value = 50 }
+        	}
+        random_country = {
+                       limit = { tag = MOR OR = { government = hms_government government = absolute_monarchy } }
```
settintotrieste, line 790: Please insert a check for is_vassal = no.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+                       relation = { who = THIS value = 50 }
+        	}
+        random_country = {
+                       limit = { tag = TUN OR = { government = hms_government government = absolute_monarchy } }
```
settintotrieste, line 794: Please insert a check for is_vassal = no.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+                       relation = { who = THIS value = 50 }
+        	}
+        random_country = {
+                       limit = { tag = LEB government = democracy  }
```
settintotrieste, line 798: Please insert a check for is_vassal = no.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+        	}
+			
+		any_pop = { 
+			limit = { strata = rich }
+			militancy = 8
+		}
+		
+		remove_country_modifier = oil_cartel
+		
+		ai_chance = { factor = 0.8 }
```
settintotrieste, line 843: Event option above already has an ai_chance of 0.8 so it should be 0.1 here

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+
+	option = {
+		name = "Greenlight the Italian Operation"
+		random_country = {
+                       limit = { tag = ITA is_vassal = no }
+                       relation = { who = THIS value = 35 }
+        	}
+			
+        set_global_flag = San_Marino_Occupied
+		ai_chance = { factor = 0.6 }
```
settintotrieste, line 1006: Option with highest ai_chance should be ordered first in the list of options

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+
+	option = {
+		name = "Withdraw from San Marino"
+		prestige = -25
+		random_country = {
+                       limit = { tag = USA government = democracy is_vassal = no }
+                       relation = { who = THIS value = -50 }
+        	}
+        743 = { add_province_modifier = { name = generic_uprising duration = 365 } }
+		ai_chance = { factor = 0.8 }
```
settintotrieste, line 1043: Option with highest ai_chance should be ordered first in the list of options

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		tag = ITA
+		year = 1957
+		NOT = { year = 1992 }
+		
+		government = democracy
+		is_vassal = no
+		war = no
+		
+		exists = RUS
+		RUS = { government = proletarian_dictatorship }
```
settintotrieste, line 1067: Should be this instead: RUS = { exists = yes government = proletarian_dictatorship is_greater_power = yes }

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		
+		HUN = { OR = { government = proletarian_dictatorship1 government = proletarian_dictatorship } }
```
settintotrieste, line 1069: Should be this instead: HUN = { exists = yes OR = { government = proletarian_dictatorship1 government = proletarian_dictatorship } }

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+			limit = { strata = middle }
+			ideology = { factor = 0.05 value = socialist }
+			ideology = { factor = -0.05 value = communist }
+		}
+
+	}
+	
+}
+
+#Italian Economic Miracle
```
settintotrieste, line 1095: Please remove this event and restore the decision. Flag check here can be used in the decision.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		OR = { has_country_flag = 1953_ITA_Centrism has_country_flag = 1953_ITA_Legge_Truffa has_country_flag = 1953_ITA_FDP }
+	}
+
+	mean_time_to_happen = { months = 2 }
+
+	option = {
+		name = "Centre-Left Coalition (DC-PSDI)"
+		ruling_party_ideology = big_tent
+		any_pop = { 
+			ideology = { factor = -0.30 value = conservative }
```
settintotrieste, line 1164: Please do not remove ideology support as that causes crashes.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		remove_country_modifier = visionary_thinker
+		remove_country_modifier = competent_placeholder
+		remove_country_modifier = imperious_autocrat
+		remove_country_modifier = prince_of_terror
+		remove_country_modifier = efficient_sociopath
+		remove_country_modifier = pig_headed_isolationist
+		remove_country_modifier = ambitious_union_boss
+		remove_country_modifier = dour_economist
+		
+		add_country_modifier = landed_gentleman
```
settintotrieste, line 1305: This is wrong, please see above

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		remove_country_modifier = visionary_thinker
+		remove_country_modifier = competent_placeholder
+		remove_country_modifier = imperious_autocrat
+		remove_country_modifier = prince_of_terror
+		remove_country_modifier = efficient_sociopath
+		remove_country_modifier = pig_headed_isolationist
+		remove_country_modifier = ambitious_union_boss
+		remove_country_modifier = dour_economist
+		
+		add_country_modifier = man_of_the_people
```
settintotrieste, line 1361: This is wrong, please see above

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+
+	trigger = {
+		tag = ITA
+		year = 1960  
+		NOT = { year = 1963 }
+		
+		government = democracy
+		
+		ruling_party_ideology = conservative
+
```
settintotrieste, line 1402: I suppose this event relates to Genoa so please insert a check for ITA owning Genoa.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		remove_country_modifier = visionary_thinker
+		remove_country_modifier = competent_placeholder
+		remove_country_modifier = imperious_autocrat
+		remove_country_modifier = prince_of_terror
+		remove_country_modifier = efficient_sociopath
+		remove_country_modifier = pig_headed_isolationist
+		remove_country_modifier = ambitious_union_boss
+		remove_country_modifier = dour_economist
+		
+		add_country_modifier = great_reformer
```
settintotrieste, line 1464: This is wrong, please see above

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+		remove_country_modifier = prince_of_terror
+		remove_country_modifier = efficient_sociopath
+		remove_country_modifier = pig_headed_isolationist
+		remove_country_modifier = ambitious_union_boss
+		remove_country_modifier = dour_economist
+		
+		add_country_modifier = great_reformer 
+		
+		any_pop = { 
+			ideology = { factor = -0.05 value = conservative }
```
settintotrieste, line 1467: No negative ideology as it will cause crashes.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1043 @@ country_event = {
...
+			scaled_militancy = { factor = 4 ideology = communist_social }
+			scaled_militancy = { factor = 2 ideology = socialist }
+		}
+		
+		remove_country_modifier = political_strife
+		add_country_modifier = { name = political_strife duration = 365 }
+		
+		set_country_flag = ITA_Democratic_Renovation
+		
+		ai_chance = { factor = 0.9 }
```
settintotrieste, line 1489: Event option with highest chance should be ordered first in the event's list of options.

settintotrieste (changes requested):

CWE/decisions/Dirigisme.txt
```diff
@@ -24,31 +24,3 @@ add_country_modifier = { name = dirigisme duration = 7300 }
 			}
 			ai_will_do = { factor = 1 }
 	}
-
-italy_dirigisme = {
```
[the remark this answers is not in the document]
Eeillios: I based it on the USSR reformation/survival events as the fate of the economy in this case would be dependent on actions taken by the player

CWE/events/italy.txt
Eeillios: I didn't put it because OTL Mattei ane ENI became close to the king if Jordan regardless of the protectorate, should I maybe use OR = { is_vassal = no is_vassal = yes)?

CWE/events/italy.txt
Eeillios: See above

CWE/events/italy.txt
Eeillios: See above, also in this case Morocco never gets free before the 60's

CWE/events/italy.txt
Eeillios: See above

CWE/events/italy.txt
Eeillios: See above

CWE/events/italy.txt
Eeillios: I thought the Hungarian Revolution flag already ensured HUN existed

CWE/events/italy.txt
Eeillios: How do I make it so an ideology looses support then?

CWE/events/italy.txt
settintotrieste: I have a policy against allowing anyone other than the overlord to affect relations with vassals. It would make sense since foreign policy is handled by the overlord so there’s no way Italy can interact directly with a vassalised Jordan to boost relations between them.

CWE/events/italy.txt
settintotrieste: Same policy applies. There’s an event for Morocco to be independent in 1956 so there’s no need for this.

CWE/events/italy.txt
settintotrieste: Flags would persist even after annexation.

CWE/events/italy.txt
settintotrieste: Increasing the ideology of one would reduce the ideology of the others. So conservative ideology is going to be reduced even without this.

CWE/decisions/Dirigisme.txt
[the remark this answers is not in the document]
settintotrieste: Sorry I want it to be triggered by a decision so the player knows which options to choose when the events pop up.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1450 @@ country_event = {
...
+		
+		FROM = { relation = { who = THIS value = -100 } }
+		
+		FROM = { 
+		    war = {							
+			    target = ITA
+			    attacker_goal = { casus_belli = acquire_any_state state_province_id = 735 }
+			    defender_goal = { casus_belli = status_quo  } 
+			    call_ally = no
+		    }
```
settintotrieste, line 1790: The event that triggers this event chain must check that Austria is independent as this event triggers a war with Italy.

CWE/events/italy.txt
```diff
@@ -455,6 +455,1450 @@ country_event = {
...
+				scaled_militancy = { factor = 4 ideology = nationalist }
+                scaled_consciousness = { factor = 2 ideology = nationalist }
+				militancy = 2
+				consciousness = 1
+			}
+		}
+		
+		734 = { add_province_modifier = { name = terrorist_chaos duration = 183 } } 
+		
+		random_country = { limit = { owns = 612 } country_event = 8020325 }
```
settintotrieste, line 1627: Must check to make sure the owner is AUS or ROA in case the allied occupation continues longer than expected.

settintotrieste (changes requested):

CWE/events/italy.txt
```diff
@@ -1899,6 +1903,237 @@ country_event = {
...
+		any_pop = { 
+			limit = { OR = { strata = middle strata = rich } }
+			ideology = { factor = 0.02 value = conservative }
+		}
+		
+		set_country_flag = 1968_ITA_Organic_Centre_Left
+	}
+
+}
+
```
Eeillios, line 2136: I know every historical event needs an ahistorical option, I will very soon add that but through events that will be triggered by this event

settintotrieste: I’m not a fan of a visual novel style of creating events (one option event leading to another) so I prefer if a direct option to avoid changing ruling party ideology is made here so as not to completely box in the player into accepting a centrist party.

CWE/localisation/countries.csv
```diff
@@ -280,4 +280,6 @@ RBU_absolute_monarchy;The Kingdom of Russia and Belarus;;;;;;;;;x
 RBU_hms_government;The Kingdom of Russia and Belarus;;;;;;;;;x
 RBU_hms_government1;The Kingdom of Russia and Belarus;;;;;;;;;x
 RDA;North Afghanistan;;;;;;;;;x
-RDA_ADJ;Afghan;;;;;;;;;x
\ No newline at end of file
+RDA_ADJ;Afghan;;;;;;;;;x
+TRB;Zone B;;;;;;;;;x
```
settintotrieste, line 284: Please give a more specific name rather than just “Zone B” can call it East Trieste or something but please no generic names.

CWE/events/italy.txt
```diff
@@ -1899,6 +1903,237 @@ country_event = {
...
+	trigger = {
+		tag = ITA
+		year = 1968
+		NOT = { year = 1973 }
+		
+		government = democracy
+		war = no
+		is_vassal = no
+
```
settintotrieste, line 2111: Please put a check that big tent is not already the ruling party

CWE/events/disasters.txt
```diff
@@ -1148,3 +1148,26 @@ province_event = {
 		}
 	}
 }
+
+country_event = { 
+	id = 294
+	title = "Vajont Disaster" 
+	desc = "EVT_294_DESC" 
+	picture = "ashgabat_earthquake"
```
settintotrieste, line 1156: Please create a event pic for this. My policy is that all historical events should use it’s own unique image.

CWE/events/disasters.txt
```diff
@@ -1148,3 +1148,26 @@ province_event = {
...
+		year = 1963 NOT = { year = 1992 }
+		owns = 733 # Udine
+	}
+
+	mean_time_to_happen = {
+		months = 10
+	}
+	
+	option = {
+		name = "EVT_235_A"
```
settintotrieste, line 1169: Please insert a comment for all option text so it’s easier for future modders to know what each event option is supposed to do.

CWE/events/italy.txt
```diff
@@ -1499,8 +1503,8 @@ country_event = {
 		
 		#Austria exists
 		OR = { 
-			exists = AUS 
-			exists = ROA 
+			AUS = { exists = yes is_vassal = no } 
+			ROA = { exists = yes is_vassal = no }
```
settintotrieste, line 1507: Might be good to add a check for war = no

CWE/events/italy.txt
```diff
@@ -1899,6 +1903,237 @@ country_event = {
...
+			consciousness = -4
+		}
+	}
+
+}
+
+#1963 elections Historical
+country_event = {
+	id = 8020330
+	title = "$YEAR$ Italian General Election"
```
settintotrieste, line 1940: Please insert check to make sure Italy is a parliamentary republic for all historical election events

CWE/events/trieste.txt
```diff
@@ -129,207 +129,486 @@ country_event = {
 	
 }
```
settintotrieste, line 131: Is there a catch all clean up event to remove TRB?

settintotrieste (changes requested):

CWE/events/italy.txt
Eeillios: The Problem is that OTL there was no way another way a majority could be formed, I tried to look into any other possible coalition but as much as I digged I couldn't find anything that wasn't just the same centrist coalition just with less parties

Eeillios: So I really wouldn't know what else to put there

CWE/events/trieste.txt
Eeillios: Will add it soon

CWE/events/disasters.txt
Eeillios: Totally forgot my bad

CWE/events/italy.txt
Eeillios: I might have found a way but would it be a problem if both options lead to a big_tent government and one of those two can later become of another ideology? alternatively if that's an issue I can make the path that can change ideology later not change ideology during the elections event

CWE/events/trieste.txt
settintotrieste: Thanks!

CWE/events/italy.txt
settintotrieste: I see, maybe the other option can just be to call another election (on the premise that a coalition was unable to be formed at all for some reason) then add some militancy here and there plus possibly some ideology support for the socialists. (must also add a election = no check if this is done).

Also, I would a like a trigger check for ITA being run by a big_tent gov for this event as I think the reasons why the election was so close was because the centrist gov faced difficulties so the effects would not make sense for any other party which is in power.

CWE/events/trieste.txt
settintotrieste: Please also modify the Treaty of Osimo event to annex TRE and TRB.

Eeillios: It does already, both are inherited and claims removed

CWE/events/italy.txt
Eeillios: Yeah another elections soon after is what I was thinking of doing. I don't know how to feel about the big_tent check as that would either mean that if you go with the other choice you don't get elections for that year or I should create an alternate history even choice, which I could do but I'd prefer to do after having finished historical path, so the current solution with the no ideology change could be kept as a compromise until I get on alternate history if it's alright with you

CWE/events/trieste.txt
settintotrieste: Okay that’s good but I think your changes haven’t been pushed to this PR yet

CWE/events/italy.txt
settintotrieste: It’s fine not to have an election event for that year if an AH path has been taken because of either an earlier event option or because some generic system intervened. The events should be scripted as tightly as possible because it’s going to be annoying to the player if an event pops up out of context, even more so if they have no option to avoid it.

CWE/events/trieste.txt
Eeillios: I didn't indeed, now I have

CWE/localisation/countries.csv
```diff
@@ -280,4 +280,6 @@ RBU_absolute_monarchy;The Kingdom of Russia and Belarus;;;;;;;;;x
 RBU_hms_government;The Kingdom of Russia and Belarus;;;;;;;;;x
 RBU_hms_government1;The Kingdom of Russia and Belarus;;;;;;;;;x
 RDA;North Afghanistan;;;;;;;;;x
-RDA_ADJ;Afghan;;;;;;;;;x
\ No newline at end of file
+RDA_ADJ;Afghan;;;;;;;;;x
+TRB;Triest - Zone B;;;;;;;;;x
```
settintotrieste, line 284: I think South Triest is better

settintotrieste (changes requested):

Eeillios:
I'm sorry but I don't think the ideology checks work for the elections system, as they don't take in account if in 1948 someone picks communists but stays democratic and same thing in 1953, secondly if ingame elections elect someone else it alters the whole system and lastly I thought the country flags worked already good enough

settintotrieste:
Sorry, it’s a policy of mine. The bigger the impact of the event, the more tightly scripted it has to be. If an event boxes in the player to accept only the ruling parties offered in the event without any choice to use generic systems to select a different ideology would degrade player agency and make the events fire out of context. Also, the butterfly effect has to be accounted for. This usually means that if the player wants to take the AH path then consequence is that all affected historical paths would be blocked out.

CWE/localisation/countries.csv
settintotrieste: I've decided to edit this to South Trieste once this PR is merged. I'm not comfortable with giving tags names that are specific to a particular event chain when the tag can be released outside of the chain through normal indy rebels, dismantlements, liberate country CB etc. The name must be contextual to all possible scenarios.

Eeillios: Ph yes sorry, the last chain was quite a brain drain and I totally forgot about it, I'll do it in the next push

Eeillios:
I understand your policy but I think that if the player doesn't want to be boxed he could just take the ahistorical option in the constitution that locks out all elections, and with the althis I have that in mind but I planned on accounting for all butterflies after historical was done.

settintotrieste:
It’s not good enough to block it with the constitution event unfortunately so I’m afraid my decision is final on this. I’d rather have overkill than underkill for trigger conditions.

CWE/events/italy.txt
```diff
@@ -2756,7 +2756,7 @@ country_event = {
 	trigger = {
 		tag = ITA
 		year = 1972 
-		NOT = { year = 1978 }
+		NOT = { year = 1973 }
```
settintotrieste, line 2759: Please try not to keep the start and end dates too close to each other as sometimes the MTTH doesn't work and the event fires one year later. This is usually when the player saves his game in 1972 to continue later.

CWE/events/italy.txt
```diff
@@ -482,8 +482,8 @@ country_event = {
 		
 		election = no
 		
-		#DC was the incumbent going into the election
-		OR = { ruling_party_ideology = conservative ruling_party_ideology = big_tent }
+		#DC or PCI was the incumbent going into the election
+		OR = { ruling_party_ideology = conservative ruling_party_ideology = big_tent ruling_party_ideology = communist }
```
settintotrieste, line 486: Sorry I can't allow this. As this is a historical event, the conditions must reflect the situation at the time the historical event fired which in this case was DC as the incumbent government.

CWE/events/italy.txt
```diff
@@ -3130,11 +3071,145 @@ country_event = {
...
+		}
+		any_pop = {
+			scaled_consciousness = { ideology = nationalist factor = 2 }
+			scaled_militancy = { ideology = nationalist factor = 2 }
+			scaled_consciousness = { ideology = conservative factor = 2 }
+			scaled_militancy = { ideology = conservative factor = 1 }
+		}
+		
+		set_country_flag = ITA_Worker_Statute
+		ai_chance = 0.9
```
settintotrieste, line 3193: Missing brackets

CWE/events/italy.txt
```diff
@@ -3130,11 +3071,145 @@ country_event = {
...
+			militancy = 4
+		}
+		any_pop = { 
+			limit = { strata = rich }
+			ideology = { factor = 0.01 value = nationalist }
+			ideology = { factor = 0.05 value = conservative }
+			consciousness = -4
+			militancy = -4
+		}
+		ai_chance = 0.1
```
settintotrieste, line 3212: Missing brackets

CWE/events/italy.txt
```diff
@@ -3130,11 +3071,145 @@ country_event = {
...
+	}
+  
+	immediate = { 
+		clr_country_flag = italy_hot_autumn
+		set_country_flag = ITA_Tradeunions_Federation
+	}
+
+	option = {
+		name = "Pass the law"
+		wage_and_work_hours = paid_sick_leave_hours
```
settintotrieste: Should be social_reform = X in the effects bracket and wage_and_work_hours = X in the trigger bracket

CWE/events/italy.txt
settintotrieste: Compromise solution is to check for ideology percentages. If the percentages are similar to the historical result then the event can fire.

Eeillios:
First of all thanks for the patience you had dealing with some of those events and fixing them, I think that soon the PR can be merged after I add the remaining events from 1970 to 1972. Second of all I wanted to ask if it was an issue if I merged two states (Emilia and Romagna) and create a new state out of the Campania one which includes the provinces of Reggio Calabria, Catanzaro and Potenza, as states could shift balances so I felt like I needed the authorization first. Third of all for event 8020340, it is supposed to be used both if the legislature lasts until 1973 or if early elections are called in 1972, is there any way to fire it in both cases, as I have seen triggering it doesn't work and I do not know if OR can be used with dates?

settintotrieste:
No problem and thanks for your contributions, they are very good!

Sure, you can modify the states so long as the total number of states Italy has remains the same at the end of the changes. I'm generally okay if the number of states increase but not if it decreases as the effect of the latter is more problematic due to the hardcoded 8 factory per state restriction. Please do check if the state changes would affect any CBs fired using events as any acquire state CB fired there will need to reference the new state capital of the state to avoid auto invalidation.

Okay, I've modified the flag check to get 8020340 event to fire as intended.

settintotrieste (approved):

PR #831, opened 2021-11-18
Sugar factory

Butterspekulatius:
Hello this is more or less only a small thing because I wanted to get familiar with pull requests.

In the begining I only wanted to add a sugar factory to represent the proccessing of sugar beets and while I was editing the buildings.txt and production_types.txt I noticed that some tried to organise the buildings but didn´t finished it.

I tried to organise the entries in a way it could make sense, especially if you build the factories manually.

Regarding the sugar factories.
I inculded them in the file but commited them out so they aren´t buildable. From my test games it seems the AI doesn´t build them regardless. Maybe the input or output has to be changed. I´m also unsure if fruits or grain should represent sugar beets.

If you have any feedback, questions or else you can also dm me on discord. I changed my name on the CWE server to the one here.


settintotrieste:
Hi, thanks for the contribution,

For the new factory, I think sugar beets can be represented using fruits. The AI is hardcoded to choose factories based on profitability so adding new factories should be fine so long as you balance the input/outputs to be in line with the average profitability of all factories which is set at 25%. You can use the production calculator to do this. 

For re-organisation edits, I prefer not to do any reorganisation of the files. It's just a preference of mine so I can know where everything is quickly. 

For adding commits, please use the github desktop client to do so instead of uploading files on github.com directly. This is as using the website can cause issues with corrupting the encoding language of the file, screwing with localisation and commits to the website doesn't always highlight the actual location of the changed code.

Butterspekulatius:
Ok. I used the calculator to get 25 % profitability with the sugar factory.
I added the sugar factory under bewerages, since they use sugar as an input.

Regarding the organisation. Understandable,  i should have asked before I edited it. 
I will keep a changed version on my end, since I find it easier to build factories this way.

I tried to use the GitHub Desktop but couldn´t figure it out how i upload the files. But I think I did it now.

Butterspekulatius:
Oh, i forgot. The factory has no loc because I don´t have any clue how to add it and I don´t want to touch the loc and break something.

settintotrieste:
I'm afraid this PR has been corrupted. Would it be possible if you could re clone your repo on the github desktop client then open another PR for the sugar factory? Also, would it be okay if you used notepad++ to edit the code files? Using normal note or wordpad tends to mess up the formatting.

PR #848, opened 2022-01-18
Merged Ryukyuan into Japanese

Butterspekulatius:
Since other small cultures where consolidaded I thought maybe Ryukyuan could be merged into Japanese.

While the culture has a tag attached to it I thought the Ryukyuan is small enough to be merged into Japanese. But I´m not sure what the requirements for culture consolidation are.

I didn´t remove it from the cultures file only commented it out in case it could be used again.


Outta-my-way-burrocrat:
I would also like to know what the standard is. Would merging some Karachay-Balkar, Bengali Hindu-Bengali, Kabardin-Circassian, amd Igbo-Ibo patches be acceptable?

settintotrieste:
@Butterspekulatius I've decided to keep Ryukyuan as there's a small but still active independence movement for it.

@Outta-my-way-burrocrat I think it would only be justified to merge cultures if there is no discrimination between the 2 and there is no active independence movement from that culture.

Probably not as the movement has been marginal at the national level so it is better to leave it to generic mechanics instead of creating a dedicated event for it.

PR #856, opened 2022-02-17
ITA Monarchist and Communist Paths

Eeillios:
Added events for Monarchist path, tweaked a few old events, added events for Communist Path, added Italian Civil War and tweaked 1948 communist path. All is working.

The pull is not finished yet, but I increased work and thought it was about time to share it with you, especially before I added too much stuff that maybe does not fit the guidelines or preferences.


settintotrieste:
Thanks. I need to request some changes to the way RIT works though.

1. RIT should not exist outside of any civil war or partition situation. So if ITA shares the same gov type as RIT then ITA must annex RIT. If RIT annexes ITA then it must change tag to ITA.

2. Following the Greek Civil War example, cores for special civil war nations must not be added to the map. This is to prevent the dismantle system from releasing 2 Italys even after it has been united for some time.

3. Generic name for RIT cannot be "Italian Republic" as that would also cover monarchy gov types. Perhaps "Italian State" can be used instead?

Eeillios:
1. Does that tolerate no exceptions? Having RIT work through ITA could cause quite some issues, as the Socialist Democracy events are not going to be compatible between the two paths and will probably cause issues with events firing when they aren't supposed to unless I was to use a ridicolous ammount of flags and checks which I'm not sure how to deal with. I also feel like it's a waste to create a tag and use it only for the Civil War and I wanted to use the tag for a few other paths coming from ITA, for the same reasons I deacribed above about ridicolous ammount of flags and checks.
2. Couldn't that be fixed by adding a string of code like, any_country = { any_owned = { limit = { is_core = ITA } remove_core = ITA } }?
3. Could changing the name for Monarchies to Kingdom of Italy fix it?

Eeillios:
I'll remove the cores and change the name, but what's the issue RIT can cause? It exists only as communist so can't mess with event as no nation has events tied to communist italy, if the event to remove all cores can in some way not remove all cores couldn't a double tag switch do it? RIT to ITA and then back to RIT so all ITA cores are removed

settintotrieste:
1. If ITA cores are removed then if RIT becomes non-communist, a catch all event needs be added to change tag back to ITA and re add the cores.
2. Only 1 union culture tag is allowed so if ITA does not exist then Italian rebels will appear for RIT to form ITA.
3. If there was a partitioned Italy where there were both ITA and RIT tags but both were subsequently annexed (by say the USSR) which was subsequently dismantled then the dismantle system might release RIT (as a mandate of one GP) and ITA (as a mandate of another GP). If RIT and ITA have cores over all their territory then there might be border gore.
4. If ITA annexes RIT then RIT cores need to removed and ITA cores re-added so another event is needed.
5. Some events reference ITA generally and should also apply to RIT.
6. If there was a partial annexation of RIT by ITA then RIT separatist rebels will spawn in ITA despite the pops not being communist.
7. Generic ICJ systems might allow ITA and RIT to demand the return of all cored lands.

There are probably more complications but these are the ones most likely to happen. All these problems can be avoided by simply preventing special tags for only 1 gov type from existing outside partition or war situations and making sure the special tag has no cores in any province at all times.

Eeillios:
Hmhhh if I don't add any core when it spawns as you suggested only 1. 2. and 5. remain as issues. for 1 an event can be done no big issue, tag_switch to a nation with no cores turns all your cores into that nation's cores, for 2 the union tag can be removed, many culture groups don't have one, shouldn't be much of an issue. For 5 I'm not sure what events are those but I can search ITA in all files and add the thing to work with both tags. Anyways I have said all I could say in defense of the tag so whatever you decide after this message I shall go by it.

settintotrieste:
(2) a union tag for ITA is needed as it has quite a few breakaway states and (5) is best not done via a mass edit so its better to keep RIT's use restricted to only partition and civil war situations.

Eeillios:
Pull is ready to merge

aurreko:
Hi [Eeillios](https://github.com/Eeillios) could you kindly upload your files as a zip archive? I'd like to test your edit, it looks interesting.

settintotrieste:
Sorry @Eeillios for the long delay on this PR. I've been unexpectedly busy this month but I should be free to start clearing up the backlog once May starts.

PR #874, opened 2022-04-24
Updated the transition to dictatorships with stricter and clearer requirements

DoktorKure:
I added stricter and clearer requirements for which type of dictatorship is ushered in when the vote franchise is set to none_voting based on ruling party ideology, minority reforms, and citizenship/war policies. This will lead to the player having more control over whether they get a sectarian regime, nationalist dictatorship, or presidential dictatorship.

### DIFF
--- a/CWE/events/clean_up.txt
+++ b/CWE/events/clean_up.txt
@@ -45,7 +45,22 @@ country_event = {
 		}
 		civilized = yes
 		vote_franschise = none_voting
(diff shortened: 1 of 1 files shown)


DoktorKure:
[governments.txt](https://github.com/settintotrieste/Victoria-2-Cold-War-Enhancement-Mod-CWE/files/8550316/governments.txt)
I also tweaked this file so populists can be part of sectarian regimes

PR #878, opened 2022-04-25
updated dictatorships through revoking vote reform

DoktorKure:
- Traditionalists and Conservatives will bring back the (absolute) monarchy if it was recently abolished
- Traditionalists and Conservatives in a country with no monarchists will usher in theocracy but only if they're moralists
- Socialists, Nationalists, Conservatives, and Populists will usher in a nationalist dictatorship if they're jingoist and not hostile to minorities
- Socialists, Nationalists, and Populists will usher in a sectarian regime if they're hostile to minorities
- Conservatives, Socialists, Populists, and Centrists will usher in a presidential dictatorship if they're not jingoist and not hostile to minorities


DoktorKure (commented):
updated for conservatives to be part of sectarian regimes to prevent issues with conservatives being jingoist AND hostile to minorities

PR #881, opened 2022-04-28
Bugfix for appointing new parties having option overflow

DoktorKure:
As it stood before, the event had an option overflow so that you would have to drag the popup box up to get more options available and even then one ideology remained off the screen (nationalists) so this addresses that issue by having the event give three options that branch off into their own events that give the choice on ruling party ideology:
status quo -> choose liberals, conservatives, centrists
leftwing -> choose socialists, progressives, communists, market communists
rightwing -> choose traditionalists, nationalists, populists)

Big thanks to @ccfox for this


Outta-my-way-burrocrat:
Why not organize them by voting coalition instead? There's only six of them right now.

DoktorKure:
> Why not organize them by voting coalition instead? There's only six of them right now.

Didn't want to make too many new events, but I'm interested in hearing which voting coalitions they are. I know there's nationalist and populists in one and there's leftwing and rightwing coalitions but that's about the limit of my knowledge of the intricacies of the voting system in CWE.

Outta-my-way-burrocrat:
There's centre for Moderates; centre_left for Progressives; right for Conservatives and Traditionalists; centre_right for Liberals; left for Socialists, Communists and Market Socialists; and minor for Populists and Nationalists.

settintotrieste:
The events for both constitutional monarchies and republics should not be merged as they are supposed to fire separate events 6016818 and 6016819

ccfox:
@Outta-my-way-burrocrat Yes, that defeats the purpose of trying to counter the overflow a little bit. The whole purpose was to reduce the number of event options and the 'voting blocs' are not made according to real life politics or coalition but tailored to fit the game. I'm pretty sure its to prevent wonky interactions between ideologies since half of them are in ideology groups that consist of a single ideology.

@settintotrieste The event could probably be edited to 'dynamically' check whether the nation is a monarchy or republic similar to some other events maybe? 

If not, to fix it Kure will just be able to take the events I made, copy-paste them, change the IDs, and alter the events fired by each option from 6016819, to 6016818.

settintotrieste:
It won’t work to use the same event. The localisation for the monarchy one is separate because MONARCHTITLE for hms gov is Prime Minister and not king like in vanilla.

DoktorKure:
Alright, time to make new commits then

DoktorKure:
All done @settintotrieste, should be good now

settintotrieste:
I think the event ID should be less than 9 digits as I’ve had issues with the game not reading such IDs.

DoktorKure:
@settintotrieste should be good now (take two)

PR #882, opened 2022-04-28
Update Revolution Spread Event.txt

DoktorKure:
updated the event for spreading your ideology in border provinces of neighboring nations since 50% chance of being discovered is actually 100% chance of discovery for some strange reason so I slightly reduced the discovery chance

### DIFF
--- a/CWE/events/Revolution Spread Event.txt
+++ b/CWE/events/Revolution Spread Event.txt
@@ -100,9 +100,9 @@ government = proletarian_dictatorship1 }
 					}
 				}
 				owner = {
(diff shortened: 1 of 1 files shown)


Outta-my-way-burrocrat:
`random` always fires no matter what. `random_list` with two or more options kinda works i think. I can't find any example of `random` in the vanilla game.

DoktorKure:
It works better yeah, though as evidenced by the space attempts being stuck at satellites because greater than 50% chance of failure, it seems there's no real random in this game at times.

PR #885, opened 2022-04-28
Create Toggle Superstates.txt

DoktorKure:
Allow the player to choose whether or not to disable the creation of post-2030s superstates, though historical superstates like the UAR will be unaffected for the sake of maintaining historical cohesion.
Note: the events for the first steps of creating the EU will still fire like normal, but the final step of forming the EU will be prevented.

### DIFF
--- a/CWE/events/Toggle Superstates.txt
+++ b/CWE/events/Toggle Superstates.txt
@@ -1,0 +1,31 @@
+#Toggle Superstates
+#Enable or disable post-2030s superstate formation in the mod
+#Activate via opening the console and keying in 'event 669967'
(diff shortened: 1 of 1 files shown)


ccfox:
Since there are some super-states that either temporarily formed historically or otherwise had a high chance to form (such as regional federations in Africa), it'd be good if you included a global flag for turning super-states off that could be used to influence the outcome of certain events via a modifier to make sure the player does not face any 'surprise' super-states.

settintotrieste:
This event won’t work as there’s a decision that will automatically clear the hide superstate flag should the country be AI run

PR #886, opened 2022-04-28
Update Nuclear Arms Development Mechanics.txt

DoktorKure:
shift base infamy away from the decision to make the bomb over to the finished development of the bomb


settintotrieste:
Infamy should come from starting development as that’s what happened to Iran and North Korea.

PR #900, opened 2022-05-06
Pop growth tweak

ccfox:
Given that there seems to be issues with nations getting quite a bit 'overpopulated' in ridiculously short amounts of time, I have made some potential changes to the natalism modifier and decision - as well as some other modifiers that gave pop growth.


settintotrieste:
Alternate system created.

ccfox:
That's far more comprehensive than what I had here, so that works way better!

PR #902, opened 2022-05-07
Update China.txt

DoktorKure:
change communist dictatorship color for China


settintotrieste:
The current colour looks better

PR #903, opened 2022-05-08
Update CAT - Catalonia.txt

ccfox:
Edited accepted cultures to match Spain.

### DIFF
--- a/CWE/history/countries/CAT - Catalonia.txt
+++ b/CWE/history/countries/CAT - Catalonia.txt
@@ -1,6 +1,6 @@
 capital = 498
 primary_culture = catalan
-culture = white_spanish
+culture = spanish
 religion = secularism
 government = hms_government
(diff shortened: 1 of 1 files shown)


Outta-my-way-burrocrat:
Are you going to remove the white_spanish culture from other places as well? It only appears in Spain 1992,

ccfox:
@Outta-my-way-burrocrat Is it in other places that it shouldn't be? Its probably in Spain for 1992 to represent immigrants from South and Central America.

Outta-my-way-burrocrat:
xac_form_union and xna_form_union still make white_spanish an accepted culture and the pink_tide decision checks if the country is primary white_spanish. Several superstates still have white_spanish as primary or accepted culture in their history files.

ccfox:
I'll take a look at them.

Outta-my-way-burrocrat:
Thank you.

PR #905, opened 2022-05-08
Goan -> Marathi

ccfox:
An opm orphaned culture that makes more sense as an existing one.

### DIFF
--- a/CWE/history/pops/1946.1.1/Portugal.txt
+++ b/CWE/history/pops/1946.1.1/Portugal.txt
@@ -1,115 +1,115 @@
 #Goa - Goa (609000/152250 POPS)
 1303 = {
     officers = {
-        culture = goan
+        culture = marathi
         religion = hindu
         size = 320
     }
 
     officers = {
-        culture = goan
+        culture = marathi
         religion = secularism
         size = 7
     }
 
     soldiers = {
-        culture = goan
+        culture = marathi
         religion = hindu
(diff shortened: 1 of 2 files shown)


Outta-my-way-burrocrat:
There are two more Goan pops in 1992.

settintotrieste:
Goan is a distinct culture of its own from any other Indian culture so it should remain. 

Wiki says that Portuguese is spoken as a first language among Goans and they don’t really mix with non Goans.

https://en.wikipedia.org/wiki/Goans

ccfox:
@settintotrieste Alright, though they are not used in any events, decisions, nations, or the like. 

Also according to the numbers on that wiki they constitute a minority of the population there in Goa now and even before the Portuguese language was not widespread among them (with the majority not speaking it as a first language). It seems as well that most of the Portuguese speakers left for other areas after annexation to India, and that currently a vast majority of the city is non-Goan altogether (1.4 million, with only 0.4 million being Goan). I still think they'd be better off merged with other cultures - even if it means dividing the city between Portuguese pops and Marathi ones.

settintotrieste:
I think it’s better to leave them there and the  create a special Goa tag for them, something like Macanese for Macau. It would also help cater for an AH situation where Goa remains under Portuguese rule much longer until it becomes more like a Macau.

ccfox:
Alright.

PR #908, opened 2022-05-10
Fixed Afghan party names

Outta-my-way-burrocrat:


ccfox:
If you want to go through nwo_parties.csv - there should be a lot of parties in there that can be removed. It actually adds more bloat/filesize to the mod having parties in the localization than it does having them in the common/countries file due to how things work (unless my maths are off).

Outta-my-way-burrocrat:
I agree. There's also QNG and UCH tags that aren't used at all; NWO got rid of them.

PR #910, opened 2022-05-11
CHI flags

DoktorKure:
Updated the flags for sectarian and republican China as the two current flags weren't used for decades at the time of the mod's start, with the blue sky and white sun design being much more prevalent even today


DoktorKure:
ignore all the other edits, idk why github does that

PR #911, opened 2022-05-11
Update CHI flags

DoktorKure:
Updated the flags for Republican and Sectarian China (not the KMT tag) as the two current flag designs hadn't been used since the 1920s and 1910s respectively.


DoktorKure:
Ignore all the other stuff, idk why github does that when I do PRs direct from my fork

ccfox:
I think the currently used flags work a lot better since the 'Five-colored flag' flag was also a proposed flag for the PRC and a 'neutral' united China anyhow - whereas KMT iconography would probably be sidelined in a China where they were not as prominent. If anything, honestly - I'd make the nationalist flag in line with the five-colored flag.

ccfox:
Also it seems you somehow make github freak out about the event template.

settintotrieste:
The five colour flag should be the democratic China flag. The sectarian flag has to be the kmt one for democracy1 which is used for CHI in the event KMT wins the civil war.

Issue #913, opened 2022-05-13
Population Natalism decision continously being fired by ai. Missing timer effect when fired


Jamernacht99:
As said in the title. Also, hovering over the selection in `Decisions` to see the effect description crashes the game

ccfox:
I can't seem to replicate the crashes - though issues with it not having a timer can certainly be fixed.

settintotrieste:
Tested on my end and there is no issue. You must have forgotten to copy and replace the event_modifiers.txt file that was edited along with the population_growth.txt file.

PR #914, opened 2022-05-13
Fix for Natalism

ccfox:
I couldn't replicate the crash mentioned in #913, however I added in a timer just to stop the AI from being able to immediately take the decision again right after the pop boost modifier expires.


settintotrieste:
Tested on my end and there is no issue. Person who reported the issue must have forgotten to update the event_modifiers.txt file

PR #916, opened 2022-05-16
Map and City Names fixes

Eeillios:
Fixed the map to represent eastern europe more accurately and changes names of cities adding events and decisions to change them throughout the years to show the historically accurate ones and redid some flags and fixed related government type to match with 1992 startdate


ccfox:
I believe there were some wonky things with this pull request given that it seems to have undone quite a few changes made since then - like Gagauzia's parties.

settintotrieste:
I'm afraid that this PR would need to be resubmitted as there are quite a lot of wonky things going on here.

PR #919, opened 2022-05-17
Update nigeria.txt

DoktorKure:
Gave Britain an option to divide Nigeria upon independence, updated the Biafra war to require the military coup, updated Boko Haram event, updated hypothetical civil war event, eliminated unnecessary event for Biafra upon independence


DoktorKure:
Let me know if you see any mistakes and I will correct them if needed

PR #922, opened 2022-05-17
Update nigeria.txt

DoktorKure:
tweaked civil war event

### DIFF
--- a/CWE/events/nigeria.txt
+++ b/CWE/events/nigeria.txt
@@ -319,11 +319,12 @@ country_event = { # Boko Haram
 	trigger = {
 		tag = NGE
 		year = 2014
-		NOT = { year = 2024 }
 		
 		NOT = { government = theocracy }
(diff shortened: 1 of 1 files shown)


ccfox:
You should make a separate tag for Hausaland/Sokoto/Northern Nigeria instead of using the ISI tag. ISI and Boko Haram are separate things, with one being a global jihadist movement and the other being a wahhabist movement within Nigeria that is also tied to ethnic separatism. 

This also will prevent issues with the ISI tag's events (which are mostly broken from vanilla NWO), 'blocks' to event chains that use the ISI tag, and issues with there being no specific separatist tag for Northern Nigeria. There are several other groups beyond Boko Haram that advocate for its secession, and it was also administered separately during colonial times.

settintotrieste:
Agree with @ccfox. ISI should not be used as much as possible and its better to use a separate tag to represent the civil war factions.

PR #923, opened 2022-05-17
Northern Nigeria

ccfox:
This tag should be used for Northern Nigeria instead of ISI. I don't have the time currently (my apologies) to add in its cores. It can also be used to model the general separatism in the North.


CWE/history/countries/NNG - Northern Nigeria.txt
```diff
@@ -0,0 +1,96 @@
...
+	education = self_financed_education
+	health_care = no_health_care
+	security_reform = police
+	child_care = no_child_care
+	unemployment_subsidies = no_subsidies
+	pollution = no_controls
+	wage_and_work_hours = unregulated_hours
+	social_service_reform = no_social_service
+	# Technologies
+oob = "1992/NGE_oob.txt"
```
settintotrieste, line 95: Cannot have OOB entry in countries not existing at the start as that can cause a crash when the player clicks on the release nation tab if this country is a possible vassal of THIS.

ccfox: I swear github just does whatever it wants sometimes. Its like HAL from 2001.

PR #925, opened 2022-05-18
Fixing the dry year picture

ccfox:
After much fuss, I have replaced the odd picture with a less odd picture.


settintotrieste:
What was wrong with the previous image?

ccfox:
The previous image is from a Somali food crisis, and I think that given that famines, droughts, and the like are capable of happening outside of Africa that the image should be more neutral so that nations like Romania, Afghanistan, India, and the like getting the event doesn't always display an image from Africa. It just seemed a bit weird.

settintotrieste:
I've decided to keep the existing image as the one proposed appears to be rather blurry. Having a brownish colour is also best avoided as it will blend in with the event UI box.

PR #926, opened 2022-05-20
Update Essential Events.txt

DoktorKure:
- Removed the chance for France to keep their colonial empire as they choose it too often
- updated the event to also fire for the FBU
- include Cameroon in AEF tag if France choose to keep Cameroon the first time

### DIFF
--- a/CWE/events/Essential Events.txt
+++ b/CWE/events/Essential Events.txt
@@ -661,7 +661,7 @@ country_event = {
 	fire_only_once = yes
 
 	trigger = {
-		tag = FRA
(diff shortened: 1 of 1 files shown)


settintotrieste:
I’m afraid I have to reject this PR:

1. Historical events cannot have options with 0 ai chance because of the no historical railroading policy
2. FBU should not receive any historical events for France as FBU is an AH situation.
3. Cameroon cannot be aggregated together with AEF as it is an international mandate and not a colony so France had no right to unilaterally merge it with other colonies.

PR #927, opened 2022-05-20
New Chinese flags

DoktorKure:
new flags for republican, theocratic, sectarian, and monarchist China based on the 5-colors flag scheme that was prevalent throughout the warlord era and even into the 1946 start date


DoktorKure:
again ignore the template event update idk why github is freaking out over that

settintotrieste:
Sorry, I don't accept anymore changes to China's flags. Everyone seems to have a view of which flag is the right one so if I accept this, someone else will say it should be the previous set so I'm sticking to the current set permanently to avoid further flip flopping.

PR #934, opened 2022-05-24
Cultural switching

ccfox:
Cultures are a mess, switched around some - can't fix them all because it requires going through events/decisions. Some weird decisions, don't know why some were made.


settintotrieste:
Please don’t delete cultures from cultures.txt. It is better to keep them for backwards compatibility with older saves.

Outta-my-way-burrocrat:
Why'd you make the Ukrainians part of the Russian pan-nationalist group?

PR #952, opened 2022-07-23
Started adding credits for commit 729e948

Outta-my-way-burrocrat:


settintotrieste:
I think it is not necessary to add detailed credits because a huge number of specific systems were taken from multiple mods over the years and I've largely forgotten what each contributor specifically added.

PR #954, opened 2022-07-26
Elia

Eeillios:
Some historical fixes in gov type, parties and flags and events for POR, GUB, CPV and STP. Localization is missing but it's the next push, if there is no issue when loc will be done this is ready to be merged


CWE/events/portugal.txt
```diff
@@ -329,6 +767,252 @@ country_event = {
...
+			add_accepted_culture = black_portuguese
+			add_accepted_culture = pardo_portuguese
+		}
+
+		ai_chance = { factor = 0.5 }
+	}
+
+	option = {
+		name = "We will maintain direct rule"
+		any_owned = { limit = { is_core = ANG } add_core = POR }
```
settintotrieste, line 1004: Cores should not be added over colonies as I have a generic system planned to add cores over colonies by expelling natives and replacing them with primary culture so if a majority of pops are primary culture then a core is added.

CWE/events/Guinea-Bissau and Cape Verde.txt
```diff
@@ -0,0 +1,377 @@
+###               ###
+### Guinea-Bissau ###
+###               ###
+
+#The Union between Guinea-Bissau and Cabo Verde - GUB
+country_event = {
+  id = 8011711
+  title = "The Union between Guinea-Bissau and Cabo Verde"
```
settintotrieste, line 8: Title is too long.

“The Guinea-Bissau and Cabo Verde Union” is better

ccfox:
Changing the economic policy of the market socialist policies from state capitalism to planned_economy1 might be a good idea, now that executives can invest.

CWE/events/Essential Events 4.txt
```diff
@@ -2283,22 +2283,88 @@ country_event = {
...
 		name = "Their support will not be let down!"
-set_country_flag = franco_boost
-any_pop = { ideology = {
-	factor = 0.7
-	value = nationalist
-} }
-					}
+		set_country_flag = franco_boost
+		any_pop = {
+			limit = { AND = { is_accepted_culture = yes strata = rich } }
```
settintotrieste, line 2297: is_accepted_culture = yes doesn't include is_primary_culture = yes so OR = { is_primary_culture = yes is_accepted_culture = yes } would be better

settintotrieste:
I have to reject these flags for the following reasons:

NAM_realm.tga (Identical to another tag's flag (SAF))
NAU_realm.tga (Stars do not actual align with the real flag)
POR.tga (Colour does not match the actual flag; it appears washed out)
POR_populist.tga (Colour does not match the actual flag; it appears washed out)
TGZ.tga (Flag is not the correct historical flag; TGZ exists as sar_government so generic flag must be the current flag; see https://en.wikipedia.org/wiki/Tangier_International_Zone)

Eeillios:
So regarding NAM and TAI flags, I chose those as those countries used the sovereign's flag when they were mandatory governments, Namibia as the Trust Territory and Taiwan as the Japanese puppet, rn Taiwan has a flag from 1895 for game start which is not that accurate, but if it's an issue no problem. Regarding NAU it seems weird they do not match, I used the one it shows when it was a trust territory https://en.m.wikipedia.org/wiki/Flag_of_Nauru  . For Tangiers instead I did some digging and the flag on wikipedia is a bad replica done by someone online, various sources show this as the actual flag http://states-world.com/symbols.php/2150 https://www.worldstatesmen.org/Morocco.htm https://www.crwflags.com/FOTW/flags/ma-t_iz.html
One last thing, I see you havw removed the culture switch from the colonies, is it an issue if I add it back? It was working in game

settintotrieste:
I see but I have a policy of no identical flags no matter what to avoid any confusion in the UI so NAM_realm.tga & TAI_realm.tga should be retained as it is currently.

For NAU_realm.tga I believed what was done was that the stars were separately resized so as not to appear squashed. Unfortunately that made the stars misaligned vis-a-vis the british canton.

Alright for TGZ.tga.

I feel that the culture switch should not be added unless there's a clean up event to switch them back to the correct culture once those countries are independent as otherwise those countries will remain portuguese until the end of the game's timespan which would be unrealistic.

Eeillios:
Pull request is ready to merge, and regarding the in game portuguese flag, it is not a wrong one, the portuguese flag does not have a specifier on the shade of the colour and the colours I put in game are the second most common choice after the brighter ones, as you can see on wikipedia there are various portuguese flags that use the same shades that are now in game

Eeillios:
Given that much time has passed I begun working on some other events so please don't merge the pull in the current state

settintotrieste:
Please submit the Tanzania changes in another PR as I want to merge the Italian ones first.

settintotrieste:
I disagree with TAN being a puppet at the game’s start. According to Wikipedia they only had their first real election in 1958 so that is the time when they should be made a puppet and not earlier

CWE/localisation/text.csv
```diff
@@ -13838,16 +13838,10 @@ TAJ_populist;Patriot Party;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
...
-TAN_colonial_government;Tanganyika;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
-TAN_mandate_government;Tanganyika;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
-TAN_hms_government;Tanganyika;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
-TAN_hms_government1;Tanganyika;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
-TAN_colonial_government_ADJ;Tanganyikan;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
-TAN_mandate_government_ADJ;Tanganyikan;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
-TAN_hms_government_ADJ;Tanganyikan;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
-TAN_hms_government1_ADJ;Tanganyikan;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
-TAN_ADJ;Tanzanian;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
+TAN;Tanganyika;;;;;;;;;;;;;x,,,;;;;;;;;;;;;;;;;
```
settintotrieste, line 13841: Please don’t switch the localisation. Current name should be the base name. This will solve any fringe issues that might arise.

settintotrieste:
I've done the separation so this PR will be merged.

PR #956, opened 2022-07-28
Tag-Switching Events - First Pass

ccfox:
Trying to edit certain events to allow you to easily play as the nations without having to save and reload.


settintotrieste:
Sorry, I don't allow these changes as it would be quite problematic to implement correctly. If players want to play as Germany or Korea they can already do so in the 1950 start which was made just for these players.

ccfox:
Alright, my apologies. I just noticed a few people complained about it often - so thought I would attempt to address it.

settintotrieste:
Don't worry I understand.

PR #958, opened 2022-07-28
Map fixes

ccfox:
Move Izmail to a Ukrainian state, and rename the state Moldova is in to reflect that Bessarabia - as an administrative region of Romania - no longer exists.

Also shortened a description that was too long.

Also I guess for some reason a quantum file for Gran Colombia is in there. It just looks like a duplicate nationalist party entry was removed.


ccfox:
Thinking on it, renaming Moldavia to 'Western Moldavia' and renaming Bessarabia to 'Eastern Moldavia' might be better than renaming it Moldova. That is, if the change is accepted at all.

CWE/common/countries/Gran Colombia.txt
```diff
@@ -146,31 +135,31 @@ party = {
 }
 unit_names = {
 	carrier = {
-		"Almirante Padilla" Buenaventura "Cartagena de Indias" Colombia "Crist�bal Col�n" "Fco. de Paula Santander" "Gran Colombia" Independiente "Nueva Granada" "Sim�n Bol�var"
+		"Almirante Padilla" Buenaventura "Cartagena de Indias" Colombia "Cristóbal Colón" "Fco. de Paula Santander" "Gran Colombia" Independiente "Nueva Granada" "Simón Bolóvar"
```
settintotrieste, line 138: I believe the encoding has been changed from ANSI TO UTF in this file. This will break the special characters when it is merged.

settintotrieste:
I’m in favour of this change. Using the Western/Eastern Moldavia name would be great.

CWE/common/countries/Gran Colombia.txt
ccfox: It probably is related to why this change is now present in every branch I try to make. I'll try to fix it, but I'm not exactly sure how it keeps putting itself back into place.

ccfox: It should be fixed now.

ccfox:
> I’m in favour of this change. Using the Western/Eastern Moldavia name would be great.

Alright, I'll change it now.

PR #959, opened 2022-07-29
Sri lanka fix

ccfox:
Due to the way the RAJ and HND tags work, it seems like removing the option to give India Sri Lanka is the best decision. Otherwise sometimes the event will not fire unless HND rebels from RAJ, resulting in odd situations.


settintotrieste:
I think the better way is just to make the event give Sri Lanka to Raj if Raj exists, if only HND exists then to HND

ccfox:
I thought about that, I assume through the 'random_country' thing? The issue would be if both exists, I don't know what the game would do...I assume give it to the first one in the list, but that might still cause weird issues.

Also, looking up things - I'm not exactly sure where it was proposed for Ceylon to be merged with the rest of India in the first place.

settintotrieste:
Yes. It can be done this way:

release = SRL # So as to ensure inherited provinces are states
random_country = { limit = { tag = HND owns = delhi NOT = { exists = RAJ } } inherit = SRL }
random_country = { limit = { tag = RAJ } inherit = SRL }

As each code is executed sequentially and random_country only applying to countries that exist, all situations should be covered. If HND is the main India (by owning Delhi) then SRL goes to them first before RAJ can inherit it. If not then RAJ inherits it. If both RAJ and HND do not exist then the default state is SRL being released without any subsequent annexation.

CWE/events/sri_lanka.txt
```diff
@@ -18,16 +18,16 @@ country_event = {
...
+		#Dominion of India			
+#		HND = { 
+#			exists = yes 
+#			is_vassal = no 
+#			OR = { 
+#				government = hms_government 
+#				government = hms_government1 
+#			} 
+#		}
+#	}
```
settintotrieste, line 30: An extra bracket was commented out

ccfox: Fixed that and altered the other things.

PR #967, opened 2022-08-18
More Austrian Changes

ccfox:
I forgot West Austria was a tag, so I fixed up those parties.

I also made election events for Austria, though not all the events I want to make are finished yet. Primarily wanted to make them because I did something different, and I might go back and change it for other nations I've done. Rather than make each election its own event, I only make significant changes in politics or significant elections their own events - with the rest being generic and having a selection of the prominent parties, with 'off-ramps' to alt-history if you continuously elect a party. It seems like it would save on events.

And finally I also made a ton of loc fixes with the characters - and '' not showing up correctly in the loc.


CWE/events/Austria - Historic.txt
```diff
@@ -0,0 +1,698 @@
+$1949 Election
```
Outta-my-way-burrocrat, line 1: typo

ccfox: Fixed, thanks.

ccfox:
Since we now have a release date for Victoria 3, I'm going to apply the election consolidation changes to the GDR, and then mark this as ready for review since I don't know if I'll finish with the rest of the events I want before then.

CWE/localisation/Chrisfox's events.csv
```diff
@@ -54,63 +54,63 @@ EVTDESC8012111;The DPRK has an interesting ideology - it incorporates elements o
...
+EVTDESC8023012107;With power in Romania now passing to anti-Stalinists, the time has come to purge the late Soviet leader's influence from within the party and Romanian society. Ana Pauker - initially seen as supportive of the direction of the new government before turning out of its favor with increasingly vocal opposition - has today announced a campaign of liberalization within the party, with even more far-reaching plans for strengthening the 'Moscow Faction' influence.;;;;;;;;;;;;;x;;;;;;;;;;;;;;;;
+EVTDESC8023012108;The 'Magyar Wing' of Romania's Communist Party was dissolved in 1953, just a year after the creation of the 'Magyar Autonomous Region' in Romania. It had provided one of the few mechanisms for Hungarians in Romania to fight against the increasing policy of 'Romanianization' from the Nationalists and Stalinists that had helped prop up the rule of Gheorghiu-Dej. \n\nGiven that the Stalinists no longer wield much political influence - the question has been brought up by more liberal-minded party members as to a potential revival of this institution.;;;;;;;;;;;;;x;;;;;;;;;;;;;;;;
```
settintotrieste, line 64: I think the localisation file has been encoded wrongly. It should be set as ANSI.

CWE/events/Austria - Historic.txt
```diff
@@ -0,0 +1,698 @@
...
+		}
+	}
+
+	option = {
+		name = "Social Democrats manage to gain majority over their coalition."
+
+		ruling_party_ideology = socialist
+
+		any_pop = { 
+			ideology = { factor = 0.2 value = socialist }
```
settintotrieste, line 58: Effects adding socialist ideology should exclude rich strata.

CWE/localisation/Chrisfox's events.csv
ccfox: That's strange, opening the file in Notepad++ says that its already in ANSI, even when I download a 'fresh' version of the file.

CWE/events/Austria - Historic.txt
ccfox: Alright, that should all be fixed now.

PR #970, opened 2022-08-20
Tech schools

ccfox:
Said I was going to make these ages ago, now finally have. Not fully implemented. There needs to be generic ones, as well as consolidated alternative ones for the USSR and clean-up events. However, maybe putting them in there will get people to do stuff with them.


settintotrieste:
I think this would be too difficult to add in properly before Oct 25.

ccfox:
Fair.

PR #971, opened 2022-08-20
Removed unnecessary ruling party change

Outta-my-way-burrocrat:


ccfox:
I'd say keep it in just to prevent any weird edge-case thing from happening - since the gov type has elections.

ccfox:
Actually that entire event could be slated for 'modernization', for some reason its using variables instead of just setting itself as 'fire_only_once = yes'.

PR #962, opened 2022-08-10
CNG, ROC, RWA and BRN content

Eeillios:
-Reviewed Congo Crisis to include the Simba rebellion and have a more fleshed out start
-Historical events for Congo-Brazzaville
-Completely redone RWA and BRN to be more historical, first of all they now start under direct control of Belgium, then there is many years of events revolving around the Hutu-Tutsi dynamics
It's quite a load of content so take your time, localization, tweaks and pictures have yet to be added

Issue #969, opened 2022-08-19
Starving aristocrats and colonial states


Outta-my-way-burrocrat:
Since the changes made to the aristocrat pop didn't change the demographics of most states without factories, many aristocrats start without any money.

Outta-my-way-burrocrat:
There are still aristocrats in colonial states

settintotrieste:
Okay, I've edited the starting decision to automatically convert executives in colonies to some other pop.

PR #975, opened 2022-08-23
Less starving aristocrats, more capitalists

Outta-my-way-burrocrat:
Mitigates the starving Executives problem by switching the Investor and Executive pops in history/pops. This merge conflicts with Elia's PR #962 but I can fix it if their PR is merged first.

### DIFF
--- a/CWE/history/pops/1946.1.1/Afghanistan.txt
+++ b/CWE/history/pops/1946.1.1/Afghanistan.txt
@@ -10,12 +10,12 @@
 		religion =  sunni 
 		size = 5288
 	 } 
-	aristocrats = {
+	capitalists = {
 		culture = pashtun
 		religion =  sunni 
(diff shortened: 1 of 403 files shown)


settintotrieste:
issue #969

ccfox:
Switching out all those pops wouldn't have a negative impact on the changes to aristocrats themselves, I assume? The fact that for communist nations, they are the 'investors'?

settintotrieste:
There should be no issue as an event will auto fire for any communist nation to demote all capitalists.

settintotrieste:
Okay, I've edited the auto demote event for communists to convert capitalists to executives so any change here would not lead to issues as any executives converted to capitalists will re-convert back to executives. This change should also help to preserve the investment pool from being wiped out after a communist revolution as the money will remain available for factory construction through executives.

PR #998, opened 2022-10-01
Remove anachronistic Jabal Druze and Alawite States from 1946 start

frogfoot9013:


CWE/history/countries/SYR - Syria.txt
```diff
@@ -2,7 +2,7 @@ capital = 902
 primary_culture = arab_levantine
 culture = arab
 religion = secularism
-government = mandate_government
+government = democracy
```
settintotrieste, line 5: Syria should start as a mandate in 1946. They only became a republic on independence.

frogfoot9013: Alright, should the flag for Syria a as a Mandatory Government be changed to the green-white-black one, because from what I can find, that was the official flag of Syria as a Mandate by 1946?

settintotrieste: Alright, the flag can be changed.

PR #1003, opened 2022-10-31
Fixed map regression in Lesser Antilles

Outta-my-way-burrocrat:
Sorry to PR this so late, I just found the regression.


settintotrieste:
No problem and thanks

PR #1004, opened 2022-11-20
Tech and invention changes 2nd attempt

Outta-my-way-burrocrat:
-Removed and replaced anachronistic inventions copied from Vic2
-Only naval_bases_tech increases maximum naval base level
-supply_range is increased by naval_bases_tech instead of ship_systems
-Naval rocket techs are moved into ship_systems
-added new naval inventions
-Added level three naval_base techs to some countries
-More late-game spacecraft techs. Currently fluff only.
-Removed carrier_battle_groups from some countries
-Added and removed third_generation_jet_fighters from some countries
I had added some fixes for the space race but something at the back of my mind says I should check if somebody else made those changes already.
Changed three culture techs

New tech pictures, and old pictures from early CWE, including missile_guidance_system, power_projection, helicopter_carrier (was sea_control_ships.tga)

bugfix

bugfix

update and bugfix

fix

fix

fix

Reorganized inventions


settintotrieste:
I'm afraid I have to reject this PR as these changes are too major to be added.

Outta-my-way-burrocrat:
@settintotrieste Is the tech PR still unacceptable? What parts need to be changed before it would be acceptable?

Outta-my-way-burrocrat:
@settintotrieste carrier_battle_groups is only unlocked in 1995, can that change go in?

settintotrieste:
Yes that can be added in as well.

PR #1011, opened 2023-01-31
Fixed ai_chance bugs in usa_elections.txt

Outta-my-way-burrocrat:


Outta-my-way-burrocrat:
@settintotrieste It might be a good idea to update release 15 to 15.1.

PR #1042, opened 2023-04-11
Typo Fixes

klorpa:


Outta-my-way-burrocrat:
@klorpa would you please fix FoxFork too? https://github.com/ccfox/CWE-FoxFork

Outta-my-way-burrocrat:
@ccfox

ccfox:
I ported over some of the fixes (having to do it manually).

I appreciate the help with editing the mod, @klorpa, there have been a lot of contributors, so its extremely difficult to go through even a small portion of the localization.

To note:
-In the US we use z instead of s in words like organization, localization, etc. Both ways are correct, and the pronunciation doesn't (or shouldn't) change.

In Elia's Loc:
- On line 89 I replaced "auspicating" with "anticipating", since "auspicating" is a bit outdated.

-On line 130, in regards to "assemblyw as" to "assembly as", it should be "assembly was".

-On line 149 I think "intimidatory" is better replaced with intimidation (since its in the context of intimidation tactics) rather than intimidating.

-On lines 194, 201, and 212 "Under the guidance of Moscow and help from the KGB [...]" is better replaced with "Under guidance from Moscow and with help from the KGB, [...]" as it sounds less clunky.

PR #1045, opened 2023-04-25
Delete CWE/doc directory

Outta-my-way-burrocrat:


settintotrieste:
I think it is better to keep this in case there are any useful assets for the future.

PR #1052, opened 2023-05-03
Naval base rebalance

Outta-my-way-burrocrat:
Naval base upgrades are only unlocked by naval_bases_tech
The first three tiers of naval base are available from 1946 Renamed artificial_island Fortified Naval Base; Moved its name to anti_ship_missiles anti_ship_missiles_invention, long_range_fire_tactic and speedy_maneuvering_tactic unlocked by missile_guidance_system submarine_base unlocks in 1960
Added level three naval base tech to some countries


settintotrieste:
I think it is better to reduce the naval base levels to match the existing techs than to clump 3 techs into the same unlocking year as that would make the techs less spread out over the game's timeline.

Outta-my-way-burrocrat:
@settintotrieste Why does the UK Russia and the USA get 1950s tech for free in 1946?

settintotrieste:
It’s probably a mistake on my part to let them have the extra techs.

PR #1054, opened 2023-05-06
Adapted Algeria fix from CWE-FoxFork

Outta-my-way-burrocrat:
https://github.com/ccfox/CWE-FoxFork/commit/5919550ff679f432453afa99f1ac78e179944626
https://github.com/ccfox/CWE-FoxFork/issues/60

### DIFF
--- a/CWE/decisions/Algeria.txt
+++ b/CWE/decisions/Algeria.txt
@@ -13,6 +13,7 @@ political_decisions = {
 			FAG = { exists = no capital_scope = { owned_by = ALG } }
 		}
 		allow = {
+			OR = { government = democracy government = democracy1 }
 			election = no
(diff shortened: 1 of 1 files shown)


CWE/decisions/Algeria.txt
```diff
@@ -55,11 +55,10 @@ political_decisions = {
 			tag = FAG
 		}
 		allow = {
-			OR = { government = democracy government = democracy1 }
```
settintotrieste, line 58: Only a democracy should be able to switch French Algeria into normal Algeria. This is as the latter is meant to represent a multicultural Algeria being governed by non whites which is something the whites would not allow such a transition if they were running a dictatorship.

PR #1077, opened 2023-05-19
Seychelles tweaks adapted from FoxFork

Outta-my-way-burrocrat:
https://github.com/ccfox/CWE-FoxFork/commit/ad5cb0c2aa3e9ac7fd4939d91b89c49abc3f0ca3#diff-0d600b0604aa09012fda95d59cd877317cec4a8d3173acbb3e225a9f31b934af


settintotrieste:
I think it should remain as populist_dictatorship if the ruling party is to be socialist. This is as putting a socialist party in a proletarian_dictatorship will trigger a decision to auto switch it to communist.

Outta-my-way-burrocrat:
@settintotrieste would changing the party to communist_social be acceptable?

settintotrieste:
Yes, communist_social would be fine.

PR #1078, opened 2023-05-21
Replaced sunni_arab check in sudan.txt

Outta-my-way-burrocrat:
@settintotrieste Does this break anything?


settintotrieste:
It should be fine. Thanks!

PR #1086, opened 2023-06-01
Delete Force Satellite Release.txt

Outta-my-way-burrocrat:
None of these events are used by the mod


settintotrieste:
These events may be useful for the sphere gov change system, if it is ever reworked, so it may be best to keep it.

PR #1089, opened 2023-06-06
Improved US ship names part 2

Outta-my-way-burrocrat:


Outta-my-way-burrocrat:
The commit message should've read: "Updated US secessionist state ship name lists using names copied from USA.txt"

PR #1100, opened 2023-06-24
Removed old localization

Outta-my-way-burrocrat:


settintotrieste:
I think this is not necessary as the removed localisation already does not appear in game.

PR #1109, opened 2023-07-04
Computers and telecommunications tweaks

Outta-my-way-burrocrat:


settintotrieste:
Please don’t change the order of goods listed in goods.txt as that would affect its icon.

Outta-my-way-burrocrat:
That is intended here: the change would make telecoms use the telephone sprite and communications use the radio sprite, I can change it back.

settintotrieste:
Alright.

PR #1130, opened 2023-08-16
No rubber in Salisbury

Outta-my-way-burrocrat:


Outta-my-way-burrocrat:
@settintotrieste why is there rubber in Salisbury?

settintotrieste:
@Outta-my-way-burrocrat It’s likely a bug leftover from the NWO mod as I didn’t edit the UK resources for this mod.

PR #1144, opened 2023-09-04
Update Zone B.txt

Outta-my-way-burrocrat:


Outta-my-way-burrocrat:
@settintotrieste Aosta, Friuli-Venezia Giulia, Sardinia and Venice lack a populist party before 2011

PR #1162, opened 2023-09-26
Fertilizer and rubber changes

Outta-my-way-burrocrat:
Merged plastic into rubber
Replaced plastic with fertilizer

This update works best with HPM's RGO change events

The stupid Washington fertilizer meme should be replaced with actual fertilizer factories in appropriate places.


Throwaway-acc-CWEmod:
This PR won't work because RGO workers can't be hired and fired; this code from  #1173 defines.lua could fix this.

`
    RGO_SUPPLY_DEMAND_FACTOR_HIRE_HI = 0.50,    --0.2 how fast pops are Hired when there is a high demand
    RGO_SUPPLY_DEMAND_FACTOR_HIRE_LO = 0.25,    --0.02 how fast pops are Hired when there is a medium demand
    RGO_SUPPLY_DEMAND_FACTOR_FIRE = 0.001,        --0.2 how fast pops are Fired when there is a low demand
`

PR #1215, opened 2023-12-08
a handful of fixes (now fully compatible with alice)

wxwisiasdf:


CWE/events/Essential Events 8.txt
settintotrieste, line 1: All changes to Essential Events 8.txt should be reversed as it breaks the UN mandate and intervention systems. Everything else is great. Thanks!

